PR #1, opened 2015-03-16
Environment setup

richardolsson:
Create simple base app and Docker environment to run it in.

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,0 +1,13 @@
+express = require('express');
+
+app = express();
+
+app.get('/', function(req, res) {
+    res.set('Content-Type', 'text/plain');
+    res.send('Hello, world!');
+});
+
+var server = app.listen(80, function() {
+    var addr = server.address();
+    console.log('Listening on http://%s:%s', addr.address, addr.port);
+});
--- a/app.js
+++ b/app.js
@@ -1,6 +1,30 @@
 express = require('express');
+http = require('http');
 
 app = express();
+
+api = express.Router()
+api.all(/(.*)/, function(req, res) {
+    // TODO: Verify data?
+    var options = {
+        // TODO: Handle any verb (check req.method)
+        host: process.env.API_HOST,
+        port: process.env.API_PORT,
+        path: req.path
+    };
+
+    var request = http.request(options, function(response) {
+        // TODO: Reserialize data?
+        response.setEncoding('utf8');
+        response.on('data', function(chunk) {
+            res.send(chunk);
+        });
+    });
+
+    request.end();
+});
+
+app.use('/api', api);
 
 app.get('/', function(req, res) {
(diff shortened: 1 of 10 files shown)


richardolsson:
Assigning @kesla for sanity check. Let me know. :)

app.js
```diff
@@ -0,0 +1,37 @@
...
+});
+
+app.use('/api', api);
+
+app.get('/', function(req, res) {
+    res.set('Content-Type', 'text/plain');
+    res.send('Hello, world!');
+});
+
+var server = app.listen(80, function() {
```
kesla: I much rather see that we listen to a unprotected port (8080?) and then setup something like iptables forwarding 80 to 8080 so the process can be started with fewer privileges.

kesla:
@richardolsson Would you be comfortable with setting up something like eslint for this? Enforce coding conventions & also sanity check some basics? Might especially be good in a project where we're mixing between two languages that are so different (python vs javascript).

And... Where are all the tests? :)

app.js
richardolsson: Not sure whether this is relevant in a docker context. Ultimately we are probably going to have some sort of front-end server (nginx?) so I'm not saying it's difficult to do if we do want to. In that case, lets use 8000 so it can be run alongside the platform docker container which exposes 8080 on the same (virtual) machine.

kesla: aight, even better.

richardolsson:
Linting would be great IMO. No tests yet, but we would probably treat them the same way we treat testing in zetkin-platform, putting them in a `tests` folder, running it inside the docker container with a `runtests.sh` script.

kesla:
@richardolsson I can write a sanity test for now if you like.

richardolsson:
Please do, @kesla! :)

kesla:
@richardolsson what's the standard python test framework?

kesla:
or rather, what will we use for tests? Want to make sure we write tests for our node stack that looks similar to the python tests (if possible)

richardolsson:
@kesla: Tough question. There are many, but they all work basically the same. There is unittesting built into Python, and the "frameworks" are usually either one of mocking-tools, or test runners. Don't know much about the former. For the latter you usually use [nose](https://nose.readthedocs.org/en/latest/) or [pytest](http://pytest.org). We use pytest, but in many ways they are interchangable, run the tests similarly and can discover more or less the same tests.

richardolsson:
@kesla: See [this pull request](https://github.com/zetkin/zetkin-platform/pull/7) in the zetkin-platform repo for some sample tests (functionality, testing over HTTP).

kesla:
@richardolsson See #2 for a few tweaks to this. I haven't touched the docker parts at all.

kesla:
oh, and I ended up using the test framework I'm most comfortable with...

PR #2, opened 2015-03-16
Improve

kesla:
Various improvements, most notably:

* `use strict` - enforces some best practises. Like defining variables.
* Split application & module into 2 distinct files - this is to prepare for...
* tests - tests that aren't depending on a special port or anything like that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
   "version": "0.0.0",
   "main": "app",
   "scripts": {
-    "start": "node ."
+    "start": "node bin/server.js",
+    "test": "tape tests/test.js"
   },
   "dependencies": {
     "express": "~4.12.0"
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -1,16 +1,21 @@
...
+  "author": "Richard Olsson",
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/zetkin/organizer-dashboard.git"
+  },
+  "version": "0.0.0",
+  "main": "app",
+  "scripts": {
+    "start": "node bin/server.js",
+    "test": "tape tests/test.js"
```
kesla: protip: you don't need tape globally installed for this to work - when running `npm test` it's adding `node_modules/.bin` to ENV so it's available.

PR #3, opened 2015-03-17
Set up "isomorphic" React

richardolsson:
Creates simple skeleton of a React app which runs the same rendering code on
the client and server. Please sanity check the setup, especially the node/npm
configuration.

Note that I have been trying to make sure there are as few requirements on the
developer machine as possible, running almost everything either from within the
docker container, or using the locally installed npm packages.


kesla:
LGTM.

kesla:
:dancers: :dancers:

richardolsson:
Yay!

PR #4, opened 2015-03-18
Data bootstrapping harness

richardolsson:
Before markup is rendered on the server, the necessary data must be loaded to
populate the react components. This needs to happen outside of React (which is
only a view layer, and should not handle state unrelated to views).

This PR creates a simple harness where data can be loaded (but does not
implement the loading or logic to select what to load yet). The data is passed
to React on the server, and then also included in the HTML so that it can be
passed to the client-side React as it boots up.

To cleanly include the data (and other dynamic content, e.g. the React markup)
in the HTML the PR includes a simple templating feature. The requirements are
so few and simple that it made more sense to use simple string substitution
than to install some templating package.

Please take a look at the new template feature including it's tests, as well
as the general layout of the data harness i zetkin/bootstrap/data.js.


richardolsson:
This was pair-coded together with @Bagera, but since none of us is very experienced with node yet, it would be great if you could take a look, @kesla. :)

zetkin/utils/template.js
```diff
@@ -0,0 +1,35 @@
...
+function makeFileRenderer(templatePath) {
+    var tpl = fs.readFileSync(templatePath).toString('utf-8');
+
+    return makeStringRenderer(tpl);
+}
+
+module.exports = {
+    makeStringRenderer: makeStringRenderer,
+    makeFileRenderer: makeFileRenderer
+};
```
kesla: I'd release this as a separate module.

kesla:
:dancers:

PR #5, opened 2015-03-18
Finding bootstrap data using Express router

richardolsson:
This is an example implementation of a router which resolves the data necessary
to render a particular endpoint. It augments the response object with this data
so that it is available to the React renderer. See #4 for context.


richardolsson:
I guess this is what Express calls a middleware, so maybe that would be a more suitable name than "router".

PR #11, opened 2015-07-21
App skeleton

richardolsson:
This PR creates a skeleton structure for the app, with routing, sections and
sub-sections, basic layout styling and so on.


richardolsson:
Sections, sub-sections and panes implemented in a way that makes development quite easy and fast.

PR #12, opened 2015-07-23
Flat pane structure

richardolsson:
Laying out the planes in a flat list instead of a tree, as in the early
interface prototype. This allows for better styling, and also makes the routing
much less tedious.

This actually creates a new type of routing for panes within sections, instead
of relying on react-router-component. Panes are resolved from the URL segment
by segment, with any params contained within the segment, i.e. /person:1337.

This doesn't care about react-router-component environments, which is not a big
deal since by the time this kicks in, the environment has already been 
interpreted and a path derived, so this should work regardless of whether there
is support for the history API or not.


richardolsson:
Merging right away, keeping for future reference.

PR #16, opened 2015-07-25
Person list and edit pane

richardolsson:
This PR creates person list and edit panes, fully integrated with the backend
via stores and actions.


richardolsson:
CRUD operations implemented. Does not indicate loading or manage errors though. We need to develop a loading and error handling strategy that is common for the entire app.

PR #17, opened 2015-07-25
Form framework

richardolsson:
This PR creates a simple form framework. The `Form` component is a wrapper for
any number of form fields, defined as concrete sub classes of the `InputBase`
base component.

The Form component renders the form fields in a concistent layout and manages
data entry and submission events. The `getValues()` and `getChangedValues()`
methods are convenient for working with the form output (e.g. sending only
the changed values in a PATCH request to the API).

### DIFF
--- a/src/js/components/form/Form.jsx
+++ b/src/js/components/form/Form.jsx
@@ -1,0 +1,77 @@
+import React from 'react/addons';
+
+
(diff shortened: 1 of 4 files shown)


richardolsson:
Merging right away, keeping PR for future reference.

PR #23, opened 2015-07-27
Header styling

Bagera:
Add styling to header and user menu

### DIFF
--- a/src/js/components/header/Logo.jsx
+++ b/src/js/components/header/Logo.jsx
@@ -1,12 +1,11 @@
 import React from 'react/addons';
+import { Link } from 'react-router-component';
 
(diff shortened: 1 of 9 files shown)


richardolsson:
I had a look and it looks mostly good. 

I do however miss the name/username and organization in the header on wide screens. It was there before, but you have moved it into the menu. It can be really useful to see what your active organization is even without hovering the menu so unless you disagree maybe you could move it back (and hide it on narrow screens)?

This is what it looks like for me:
image

This is from our UI concept sketch:
image

richardolsson:
Ok I'll go ahead and merge this and we can discuss the user info at a later time.

PR #26, opened 2015-07-28
Section menu styling

richardolsson:
Adds basic styling (still no icons) for the section nav menu.


Bagera:
We are mixing pixels and ems. I would prefer stickig to only ems, that way it is more accesible and wont break the layout as easily.

richardolsson:
Ok, I'm not sure I understand the reason, but I've translated all px scalars in this PR to em. :)

One of us can open a PR for project-wide conversion at a later time. Good to merge?

PR #30, opened 2015-07-29
Scoped search

richardolsson:
This implements scoped search, i.e. searching for only a single category of content, e.g. maps content, people or campaign content. It adds a widget to the search form for changing the scope and keyboard shortcuts for focusing the search field with a predefined scope.


richardolsson:
Here is a screenshot of the search field with the new (still unstyled) widget.

image

PR #31, opened 2015-07-29
Vertical (mobile) pane layout

richardolsson:
On devices with narrow screens the horizontal pane layout does not work. This PR creates a layout for narrow screens, where panes are stacked vertically and can no longer be dragged around. Scrolling in this layout is not managed, but instead relies on browser scrolling which is both easier to implement and will likely work better on devices with narrow screens (which tend to be slower).

image

The screenshot shows a (contrived) situation where six panes have been stacked on top of eachother in this new vertical layout. It will be possible to jump back in the stack by tapping the name of an underlying pane.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "test": "tape tests/*.js"
   },
   "dependencies": {
+    "animated-scrollto": "^1.1.0",
     "array.prototype.find": "^1.0.0",
     "bugsnag": "^1.6.4",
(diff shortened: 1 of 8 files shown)


richardolsson:
There is still a lot of work left before this can be merged, but what do you think about the concept, @Bagera?

richardolsson:
This is ready for review. I'm happy with the results.

Bagera:
I like it alot.

PR #32, opened 2015-07-30
Better preloading

richardolsson:
There was a lot of boilerplate involved in preloading data for every route, even though they essentially all did the exact same thing; execute one or several actions and invoke next() when it finished. The new waitForActions() function does this and makes the code much shorter. 

This PR also adds previously missing preloaders for people and locations.


richardolsson:
Feel free to take a look at this simple PR too, when you get a chance, @Bagera. :)

PR #38, opened 2015-08-01
Activity list and panes

richardolsson:
Add activity list to campaign section, and panes to add, edit and delete activities.

### DIFF
--- a/src/js/actions/activity.js
+++ b/src/js/actions/activity.js
@@ -9,6 +9,11 @@ export default class ActivityActions extends Actions {
         this.flux = flux;
     }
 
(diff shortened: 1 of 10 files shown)


richardolsson:
Merging to proceed work.

richardolsson:
`AddActivityPane` and `EditActivityPane` are both in `src/components/panes` on the master branch. The `AllActivitiesPane` seems to have been removed, probably because there was no use for it (activities are not listed anywhere except when the user selects them, since they are "secondary data" for actions). The `ActivitiyForm` is used by both of the remaining panes, and is in `src/components/forms` on master.

PR #39, opened 2015-08-03
Feature/campaign management

j0n:
Added Create, update and delete for Campaigns.


richardolsson:
Added min-width of new panes. Other than that this LGTM. :+1:

PR #72, opened 2015-08-24
Avoid npm (re-)install

richardolsson:
A developer who already has all dependencies locally shouldn't have to install
them inside the container too.


richardolsson:
I'm gonna go ahead and merge this right away. But please have a look and we can revert/change eventually if you want to go another way.

qzio:
Looks awesome!

PR #75, opened 2015-08-26
Campaign filter

richardolsson:
Feature for filtering content on campaign. Useful in the action list (display
actions only for certain campaigns) and in the campaign overview pane (only
display statistics et c for selected campaigns).


richardolsson:
Example of what it looks like in the actions list.
image

PR #76, opened 2015-08-27
Campaign calendar view

richardolsson:
Creates a component to display actions in a calendar and adds it as a second view mode to the "All actions" sub-section.

The `ActionCalendar` component is completely "dumb", displays all actions that are given to it (assuming that they are ordered correctly), allows clicking on an action to "open" (or whatever the parent component dictates) and supports drag and drop between days.

image


richardolsson:
Please have a look at this @j0n and let me know what you think. Can be developed further, but ready to merge basic functionality.

PR #122, opened 2015-09-02
Add activity breakdown next to locations

### LINKED ISSUE #82: Activities breakdown similar to "Campaign locations"
Some sort of breakdown is required. It should show the same kind of information as the "Campaign locations" section, but since there will rarely be more than 5-10 activities in a campaign (whereas there could easily be 50-100 locations) it might be possible to solve it in some other way.

It might be a good idea to have the two side by side. Mostly because the activity depends on the location and vice versa. If you conclude that you have very few actions of type X, the next step will be to look for good candidates to change from Y to X. Not all Ys are candidates though, because some could be in a location that is not suitable for X.

One could argue that this is a non-issue, since the organizer could just create a new action, or change the location too if the action they choose to edit is not in a suitable location.

richardolsson:
This generalizes the `ActionLocations` component to also support activites, or really any { title } object on actions. It uses this new `ActionDistribution` component to show both locations and activity distribution in the old "Campaign locations" pane, and renames it "Distribution".

image

Closes #82


richardolsson:
How about this, @jensborje?

richardolsson:
Merging after discussions with @jensborje.

PR #123, opened 2015-09-02
Person query framework

### LINKED ISSUE #111: Person mega-search
Create a completely separate sub-section for searching (or rather, creating queries) for people. Since almost everything in Zetkin is connected to a person, directly or indirectly, there are tons of parameters on which to filter out people.

The interface will need to allow for adding filters dynamically. A query can consist of one or many filters. It should be possible to save a query for later re-use.

Example of filters:
* Person data (name, e-mail, join date et c)
* Participation in specific action or campaign
* Participation in actions on specific time-of-day
* Participation in actions of a specific activity
* Whether responded to a given survey
* Response to a specific survey question
* Whether was called as part of specific call assignment
* Whether was called in last N unit-of-time

richardolsson:
This PR creates a framework for setting up complex person queries by chaining different kinds of filters. It creates some basic filters (some of which are noops simply for demo purpose at this point), and an `EditQueryPane` for adding and configuring filters into a query.

image

The PR also implements searching for stored queries and opening their results in a separate `QueryPane`, so that you can search for example for a cadre shortlist to more easily find people who are easy to engage in one of tomorrow's actions. Although not yet implemented in `PeopleList` (#133), it should be possible to drag people directly from this shortlist onto an action to book them.

image

Closes #111

### DIFF
--- a/src/js/actions/query.js
+++ b/src/js/actions/query.js
@@ -2,6 +2,10 @@ import { Actions } from 'flummox';
 
 
 export default class QueryActions extends Actions {
+    createQuery(title) {
+        return title;
+    }
+
     updateFilter(queryId, filterIndex, filterConfig) {
         return { queryId, filterIndex, filterConfig };
     }
--- a/src/js/actions/query.js
+++ b/src/js/actions/query.js
@@ -1,0 +1,8 @@
+import { Actions } from 'flummox';
+
+
+export default class QueryActions extends Actions {
+    updateFilter(queryId, filterIndex, filterConfig) {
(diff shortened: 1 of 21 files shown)


richardolsson:
To actually evaluate the queries is quite complex and requires either retrieving a lot of data, or being closer to the database (e.g. in the platform layer). A better solution will need to be found on how/where to perform searches like these, but for now they are just executed locally against the local list of people.

richardolsson:
Merging this to continue work. Will discuss and possibly tweak UX at Sunday's hackathon.

PR #142, opened 2015-09-03
Better user menu

### LINKED ISSUE #117: Tweak user menu
The user means needs some work. 

* Display name instead of e-mail address
* Use the `Avatar` component to render user picture
* Make the expanded menu "pop" more over the underlying content

This is what it looks like right now.

image

 There is currently no clear indication in the header that the menu is open, and since it's entirely flat (without any shading or anything to make it pop out) it could easily blend into whatever is displayed below it and missed.

richardolsson:
This PR tweaks the styling and content of the user menu.

* Show name instead of e-mail address of user
* Show real avatar using the `Avatar` component
* Fade down background when hovering menu to make it pop

usermenu

Closes #117.

### DIFF
--- a/src/js/components/header/UserMenu.jsx
+++ b/src/js/components/header/UserMenu.jsx
@@ -2,6 +2,7 @@ import React from 'react/addons';
 
 import FluxComponent from '../FluxComponent';
 import OrgPicker from './OrgPicker';
+import Avatar from '../misc/Avatar';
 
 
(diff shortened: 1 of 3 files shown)


richardolsson:
What do you think about this, @Bagera?

Bagera:
Looks good. It's better with name than mail and the avatar makes sense. I'm not that convinced about the fade but I don't have any better idea. I give the PR a thumbs up

PR #218, opened 2015-09-18
Add explicit dependency to react-async.

ramhoj:
Or gulp watch could not build js compile task.

node v0.12.7
npm 2.14.2

(Same as the PR to accounts).


richardolsson:
This is correctly observed, although `react-async` should not be required. Apparently, it's been made into a `peerDependency` for `react-router-component`, but will be removed in the latest version.

Issue #129, opened 2015-09-02
CSS refactor: Per-component styling


richardolsson:
The SCSS source is starting to get big. As it gets bigger, the need for structure and consensus gets more pressing. This is a suggestion for how we could structure SCSS going forward.

### Per-component styling
Right now it's unnecessarily difficult to guess the class name, and even more so the SCSS source location, for a particular JSX component. This is completely unnecessary, since all components already have (more or less) unique names, with corresponding file names.

I propose that the top element of every component be given a `className` that is an all lowercase version of the component name, e.g. `PersonList` becomes `personlist`. Furthermore, the SCSS source pertaining to the `.personlist` selector is placed in `PersonList.scss` _next to_ the `PersonList.jsx` component JavaScript source code. For example, the file structure might look something like this:

* `src/components/personlist/`
  * `PersonList.jsx`
  * `PersonList.scss`
  * `PersonListItem.jsx`
  * `PersonListItem.scss`

The SCSS files could be collected automatically, e.g. using [gulp-sass-glob](https://www.npmjs.com/package/gulp-sass-glob).

### Responsive breakpoints
Media queries should go in each component SCSS file, so that all the styling related to a particular component can be found in one place, as a general rule of thumb. The actual pixel widths of breakpoints should be externalized to some central location, e.g. using variables. A SitePoint article describes [some solutions for this](http://www.sitepoint.com/managing-responsive-breakpoints-sass/).

### Common styles
Some styles and SCSS assets will be shared across the entire project. Those should still go in the `src/scss` folder, along with the `style.scss`.

### Caveats
There are some risks and possible pitfalls associated with this solution. 

#### Non-deterministic concatenation
If the SCSS files are collected automatically there will be no way of pre-determining the order in which they are concatenated. That means that order can't be relied upon for overrides. This means the only other option (ruling out `!important`) is specificity.

#### Name collisions 
Although most components have globally unique names, they are only required to be unique within a folder. This means that there could (and likely will) be occurrences of components with the same name appearing in two contexts. This means components can never have generic names like `Button` or `Footer` for the risk of causing conflicts.

## SCSS style
Below is an example of what the SCSS file for `MyComponent.jsx` would look like.

```scss
// MyComponent.scss
// Single top level block
.mycomponent {
  // Base, mobile-first styling
  width: 98%;
  margin: 0 1%;

  // Elements below top level have classnames consisting of the
  // component class name with a descriptive, locally unique suffix.
[...]

PR #221, opened 2015-09-20
Refactor/component scss

Bagera:


richardolsson:
As discussed in #129.

richardolsson:
The last few commits add a new `componentClassNames()` utility function to the directory level of `src/js/components`, and implement it in `WidgetBase`. We might want to move it to a separate NPM package eventually.

The usage in `WidgetBase` is a good example of how to use it:

```javascript
import cx from 'classnames';
import { componentClassNames } from '../../';

// ...

const classes = cx(componentClassNames(this));
return <div className={ classes }/>;
```

The prototype chain is traversed up to but not including `FluxComponent` or `ReactComponent`, and an array is returned with the names of the constructors. That array can be joined using `.join()` or passed to the `classnames`/`cx()` utility, and the resulting string should be used as the value for `className` of the top-level element in the component.

@Bagera: Please use this whenever possible going forward. Even if there is no inheritance, using this utility means that if we later decide to change the inheritance chain, the new set of classes will be assigned automatically.

richardolsson:
If you don't mind, @Bagera, I will take over this PR, finish the work and merge it.

PR #238, opened 2016-05-09
Form styling

jensborje:
This PR adds styling to the Form component (and higher-order components) and input field components.


jensborje:
Closing because branch should start at latest master head.

PR #248, opened 2016-05-21
Call filters

richardolsson:
This PR creates filter UI for filters related to phone banking.


richardolsson:
More call related features may be added in the future but this is all that is needed for the assignment templates, and most other use cases.

PR #257, opened 2016-05-25
Call assignment stats styling

jensborje:
This PR adds styling to Call assignment stats of target and goals.


richardolsson:
LGTM! 🍾

PR #291, opened 2016-09-21
Interface to edit officials

richardolsson:
This PR adds a new section, "settings", and a sub-section called "Officials" where members of the organization can be promoted (or demoted) to admins or organizers.


richardolsson:
New officials can now be added by dragging them from any context and dropping them in the appropriate role, or through the `SelectPeoplePane`.

Existing officials can be moved around between roles using drag-and-drop, and removed completely using the delete icon.

Click an official opens them in the `PersonPane`.

image

PR #296, opened 2016-09-24
Location tags

jesper-lindberg:
This PR adds a new `LocationPane`, for editing tags and then some.


richardolsson:
These are the relevant endpoints:

* `/orgs/{org_id}/locations/tags` (GET to retrieve tags, POST to create new tag)
* `/orgs/{org_id}/locations/{loc_id}/tags` (GET to retrieve tags for location)
* `/orgs/{org_id}/locations/{loc_id}/tags/{tag_id}` (PUT to tag)

src/components/panes/LocationPane.jsx
```diff
@@ -0,0 +1,68 @@
...
+
+    renderPaneContent(data) {
+        if (data.locItem) {
+            console.log(data.locItem)
+            // TODO: Render form
+            return (
+                <div>
+                    <p>{ data.locItem.data.info_text }</p>
+                    <a onClick={() => {
+                        this.onLocationEdit(data.locItem.data)
```
richardolsson: This way of setting an onClick is unnecessarily complex. It could be replaced with a simple `onClick={ this.onLocationEdit.bind(this) }`, which IMO is more concise.

This does however mean that the location object is not passed to the handler function, but that doesn't need to be a problem.

src/components/panes/LocationPane.jsx
```diff
@@ -0,0 +1,68 @@
...
+                </div>
+            );
+        }
+        else {
+            // TODO: Show loading indicator?
+            return null;
+        }
+    }
+
+    onLocationEdit(loc) {
```
richardolsson: Instead of passing the location to `onLocationEdit()`, you can retrieve the current location ID using `this.getParam(0)` (it's the first param to the pane in the URL/view store).

richardolsson:
I have updated the API on dev. The aforementioned resource endpoints now exist.

src/components/panes/LocationPane.jsx
jesper-lindberg: Yep, totally agree. This will be changed to `<a onClick={ this.onLocationEdit.bind(this, data.locItem.data) }>Edit</a>` in an upcoming commit.

jesper-lindberg:
Is it correct to assume that the response from `/orgs/{org_id}/locations/` should include a property like `tags`? If so, something might be missing or I'm doing something wrong because it's not there. I've only tried adding an existing tag to a location (the PUT request seems to be successful).

src/components/panes/LocationPane.jsx
richardolsson: There should be no need to bind `data.locItem.data`. That is only necessary when the callback (`onLocationEdit()` in this case) can't know which specific location was clicked, like when there are several, e.g. in a list.

Instead, you can just use `this.getParam(0)` inside the callback (see other comment), which means that the bind only needs to contain `this`, like so:

```javascript
<a onClick={ this.onLocationEdit.bind(this) }>Edit</a>
```

richardolsson:
No, the response from `/orgs/{org_id}/locations` should be an array of location objects, and those do not have a `tags` property.

The list of all location tags are found at `/orgs/{org_id}/locations/tags` (similar pattern as with people tags, which are at `/orgs/{org_id}/people/tags`. 

The list of tags used to tag a specific location is at `/orgs/{org_id}/locations/{loc_id}/tags`. You tag by adding to that collection resource, and untag by deleting from it, using `PUT` and `DELETE` respectively, at `/orgs/{org_id}/locations/{loc_id}/tags/{tag_id}`.

So if I understand your question correctly, you're looking for `GET /orgs/{org_id}/locations/{loc_id}/tags`.

src/components/misc/loclist/LocationList.jsx
```diff
@@ -38,7 +38,7 @@ export default class LocationList extends React.Component {
                         return (
                             <LocationListItem location={ loc }
                                 key={ loc.data.id }
-                                onSelect={ this.onSelect.bind(this, loc) }/>
+                                onSelect={ this.onSelect.bind(this, loc.data) }/>
```
richardolsson: Please, if it's possible, undo this action, instead changing `components/sections/maps/LocationsPane` so that the `data` attribute is retrieved from the item object there.

That way, we can stick to a single pattern where lists dispatch the entire item (introduced in #294).

richardolsson:
LGTM

Great work! Merging this. Let me know when you want another task. :)

jesper-lindberg:
Woohoo! 🎉

PR #297, opened 2016-09-24
Action summary pane

richardolsson:
This PR adds a new `ActionPane` summary pane, to replace `EditActionPane` in most cases (except when user explicitly wants to edit the action). The `ActionPane` contains (or will at some future point contain):

* Action summary (time, location et c)
* Organizer notes for action
* List of action participants
* List of responses to be booked to action

### DIFF
--- a/src/components/lists/items/elements/ParticipantItem.jsx
+++ b/src/components/lists/items/elements/ParticipantItem.jsx
@@ -21,12 +21,13 @@ const participantSource = {
         const newAction = dropResult.newAction;
         const targetType = dropResult.targetType;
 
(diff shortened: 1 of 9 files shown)


richardolsson:
A list of current participants are now rendered in the `ActionPane`, and people can be dragged and dropped onto the action to add participants.

image

Consider instead using the same approach as in the caller list and officials section, where people can be added both by dragging, and by selecting?

richardolsson:
This new pane will need to be augmented with a few more features eventually, such as:

* Response list and booking responses as participants
* Icons to indicate status (booking status, reminders sent et c)
* Organizer notes

PR #305, opened 2016-10-10
Internationalization and Swedish strings

### LINKED ISSUE #59: Internationalization
Implement nationalization and localization using some i18n/l10n library, e.g. [L20n](http://l20n.org) or [formatjs/react-intl](http://formatjs.io/react/).

richardolsson:
This PR replaces all hard-coded strings with properly internationalized messages. It also adds Swedish localizations for all strings.

Fixes #59.


richardolsson:
More or less every component has been internationalized. There are still some bits that have been left untouched. One such element is the location pane(s). This is because they will need to be revamped anyway.

Another aspect that has been completely left out of this PR are dates and times. A proper solution for these will be necessary at some point, but the universal rendering makes it a little bit complex. If we are unable to fix this before release, it's not the end of the world.

PR #312, opened 2016-10-22
Feature/call assignments list

jesper-lindberg:
This PR adds a new list component called `CallAssignmentsList` which displays target stats, goal stats and participants for each call assignment.


src/components/lists/items/CallAssignmentsListItem.jsx
```diff
@@ -0,0 +1,71 @@
...
+
+import LoadingIndicator from '../../misc/LoadingIndicator';
+import ParticipantList from './elements/ParticipantList'
+
+import {
+    retrieveCallAssignmentStats,
+    retrieveCallAssignmentCallers,
+} from '../../../actions/callAssignment';
+
+@connect(state => state.callAssignments)
```
richardolsson: What's the purpose of this? As far as I can tell, no properties from state.callAssignments are used in the component. And even if they were, this way of mapping state to properties is a little bit weird.

PR #316, opened 2016-11-01
Bulk operations

richardolsson:
This PR adds a "bulk operation" feature to lists, creating a uniform interface for executing an operation on many objects at once, e.g. tagging several people with the same tags.

image


richardolsson:
A framework has been implemented, including list UI, modular configuration of operations in the `BulkOpPane` and actions and server-side API for executing the operations. That's enough for this PR. More ops will be added down the line.

Issue #334, opened 2017-01-05
Style call assignment list


richardolsson:
The call assignment list (at /dialog/assignments) is currently very crudely designed.

![image](https://cloud.githubusercontent.com/assets/550212/21673660/51d7758a-d32a-11e6-9083-801b75bdb259.png)

richardolsson:
Maybe something inspired by the GitHub milestone listing?

![image](https://cloud.githubusercontent.com/assets/550212/21673692/6b9e3b2a-d32a-11e6-94a1-6dcd22195a5b.png)

jensborje:
Something like this?

![skarmavbild 2017-01-11 kl 15 09 47](https://cloud.githubusercontent.com/assets/1271517/21851604/17b5b2f0-d810-11e6-99f5-15562d4ba73a.png)

richardolsson:
That's nice! Some thoughts:

* Maybe make success rate a percentage instead of number of successful calls?
* The progress bar in `CallAssignmentPane` is red, so maybe this should be to?
* The current implementation of `CallAssignmentItem` already has a list of callers, similar to the one in `ActionListItem` (in fact, using the same `ParticipantList`). Maybe keep that somehow instead of just printing number of callers?
* Maybe add a "Stats" label in the header over the pseudu-column that contains the stats?

richardolsson:
RE the caller/participant list: It would also be completely fine to remove the list of callers if we want to. In fact, that would reduce risk of bugs. I'm just thinking since it's already implemented, maybe we could keep it, but I'll let you decide.

jensborje:
I agree with the input.

I removed the caller/participant list since you assign callers in the opened pane. Having them in the list might cause confusion on where you do this, as we allow it in the ActionList. It also makes somewhat sense that the list is a summary and you interact with it do view details.

richardolsson:
Ok great!

![Make it so](https://media.giphy.com/media/wNlks0ID1igO4/giphy.gif)

jensborje:
I'm on it. Updated mockup:

![skarmavbild 2017-01-11 kl 15 30 11](https://cloud.githubusercontent.com/assets/1271517/21852319/e4b91b96-d812-11e6-9911-0856bad28069.png)

richardolsson:
[...]

PR #378, opened 2017-01-11
Call assignment list styling

jensborje:
This PR gives some styling according to proposals in #334 

skarmavbild 2017-01-11 kl 19 49 43

Solves #334

### DIFF
--- a/src/components/lists/items/CallAssignmentListItem.jsx
+++ b/src/components/lists/items/CallAssignmentListItem.jsx
@@ -43,6 +43,8 @@ export default class CallAssignmentListItem extends React.Component {
 
     render() {
         let assignment = this.props.data;
+        let callsStats = null;
+        let reachedStats = null;
         let targetStats = null;
         let goalStats = null;
         let participantIndicator = null;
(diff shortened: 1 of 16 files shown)


jensborje:
I'm unsure how the sorting works in List component.

src/components/lists/items/CallAssignmentListItem.jsx
```diff
@@ -41,48 +44,104 @@ export default class CallAssignmentListItem extends React.Component {
...
+        const inPast = (assignmentDateEnd < (new Date()) ? true : false);
+
+        const classNames = cx({
+            'CallAssignmentListItem': true,
+            'past': inPast
+        });
+
+        const assignmentDateSpan =
+            assignmentDateStart.getDate()
+            + "/" + assignmentDateStart.getMonth()
```
richardolsson: `Date.getMonth()` returns months zero-indexed, i.e. january is 0, february is 1, december is 11. So this will yield a string that is not what a human would expect.

src/components/lists/items/CallAssignmentListItem.jsx
```diff
@@ -41,48 +44,104 @@ export default class CallAssignmentListItem extends React.Component {
...
+        const assignmentDateStart = new Date(assignment.start_date);
+        const assignmentDateEnd = new Date(assignment.end_date);
+        const inPast = (assignmentDateEnd < (new Date()) ? true : false);
+
+        const classNames = cx({
+            'CallAssignmentListItem': true,
+            'past': inPast
+        });
+
+        const assignmentDateSpan =
```
richardolsson: I think we need to include year in this. I have some really long-running assignments in my dummy data, and was confused by the date output because I didn't first realize that one assignment started in 2016 and ended at the end of 2017.

src/components/misc/ProgressBar.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+export default class ProgressBar extends React.Component {
+    static propTypes = {
+        progressSum: React.PropTypes.number.isRequired,
```
richardolsson: Why "`progressSum`" and not just `progress`? This isn't really a sum (it's a fraction or a product), and it seems unnecessary to make the property name longer than necessary.

Actually, I would probably make this a fraction between 0-1, and multiply by 100 inside the component, instead of, like now, multiplying by 100 and passing the multiplied value as a property.

src/components/lists/items/CallAssignmentListItem.jsx
richardolsson: Use the date formatting utility on `Date` (added by the sugar-js library) instead, i.e. something like `assignmentDateStart.format('{d}/{M}, {yyyy}')`.

src/components/panes/CallAssignmentPane.jsx
```diff
@@ -161,10 +162,7 @@ export default class CallAssignmentPane extends PaneBase {
                         onClick={ this.onClickEditTarget.bind(this) }/>
                     <Link msgId="panes.callAssignment.target.editGoalLink"
                         onClick={ this.onClickEditGoal.bind(this) }/>
-                    <div className="CallAssignmentPane-progress">
-                        <div style={{ width: progressSum + '%' }} 
-                            className="CallAssignmentPane-progressContent"/>
-                    </div>
+                    <ProgressBar progressSum={ progressSum }/>
```
richardolsson: After having renamed the property to `progress` as discussed, this needs to updated.

PR #391, opened 2017-01-18
CallList & CallPane styling

### LINKED ISSUE #342: More information in CallPane
Currently only a minimal amount of information about a call is displayed in `CallPane`. The pane should at least contain the same information that is presented by the call log (`TargetPane`) in Zetkin Call.

Once #326 is implemented, there should also be organizer notes here. Once message from caller to organizer has been added to the API, that should also be here (and the possibility to mark call with organizer action taken).

### LINKED ISSUE #386: Improve CallList
`CallList` should be restyled to be more consistent with other dated lists in Organize. We also need to make sure that it includes the following pieces of information:

* Target person
* Caller
* Status (reached/not reached)
* Date
* Whether organizer action is needed
* Whether organizer action was taken

If we can somehow include the following without cluttering the list, that would be nice, but it's not a requirement:

* Message (truncated) to organizer
* Message (truncated) to future callers

When performing this revamp, lets try to accommodate the fact that at some point in the future, this list will not just contain calls but also other types of contact, e.g. e-mail and SMS. We should not put too much effort into this right now though.

jensborje:
skarmavbild 2017-01-19 kl 15 56 17

skarmavbild 2017-01-20 kl 16 54 21

Resolves #386 
Resolves #342

### DIFF
--- a/src/components/lists/items/CallListItem.jsx
+++ b/src/components/lists/items/CallListItem.jsx
@@ -39,21 +39,25 @@ export default class CallListItem extends React.Component {
                     <span className="time">
                         { timestamp.format('{HH}:{mm}') }</span>
                 </div>
-
-                <div className="CallListItem-target">
-                    <Avatar className="CallListItem-targetAvatar"
-                        person={ call.target }/>
-                    <span className="CallListItem-targetName">
-                        { call.target.name }</span>
+                <div className="CallListItem-content">
+                    <div className="CallListItem-target">
+                        <Avatar className="CallListItem-targetAvatar"
+                            person={ call.target }/>
+                        <span className="CallListItem-targetName">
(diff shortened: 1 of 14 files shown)


src/components/lists/items/CallListItem.jsx
```diff
@@ -11,7 +11,7 @@ export default class CallListItem extends React.Component {
 
     render() {
         let call = this.props.data;
-        let timestamp = Date.utc.create(call.allocation_time);
+        let timestamp = new Date(call.allocation_time);
```
richardolsson: Why this change? This will potentially render a date in the wrong timezone, offsetting it by one or more hours. I just noticed this so I'm commenting now even though I know this is early in the process. I just don't want to forget. Hope you don't mind.

PR #477, opened 2017-02-24
Better person I/O

richardolsson:
This PR implements the new Platform import API which is vastly superior to the middle-end implementation it replaces, both performance-wise and in the way that it handles failure (using transactions to roll back a half-baked import). The PR also re-enables the previously disabled XLSX export of persons.

Combined, this means added power to combining Zetkin with other systems.

## New ID column types
image

## Styled reports
image

## Proper error handling
image


src/components/sections/people/ImportPane.scss
```diff
@@ -32,4 +32,27 @@
             }
          }
     }
+
+    .ImportPane-error {
+        .Button {
+            @include button;
```
jensborje: Perhaps we should use inline-button() mixin instead of button(), making it less demanding for action.

src/components/sections/people/ImportPane.scss
```diff
@@ -32,4 +32,27 @@
...
+            margin: 1em 0 4em;
+        }
+
+        li {
+            font-size: 1.5em;
+            margin: 0.2em 0;
+        }
+
+        .Button {
+            @include button;
```
jensborje: Same here, using mixin inline-button() would make it more optional.

richardolsson:
Great! Have updated the button mixins. Merging.

PR #496, opened 2017-03-02
Misc reminder issues

richardolsson:
This PR fixes various minor issues related to action reminders. See commit messages for details.

### DIFF
--- a/src/components/panes/ActionReminderPane.scss
+++ b/src/components/panes/ActionReminderPane.scss
@@ -33,23 +33,19 @@
             position: relative;
             margin: 1em 0;
             padding: 0.5em;
-            height: 6em;
 
             .Avatar {
-                width: 5em;
-                height: 5em;
-                float: left;
+                @include col(2,12, $align:middle, $first:true);
+                max-height: 6em;
             }
 
             .ActionReminderPane-remindedInfo {
-                float: left;
-                margin-left: 1em;
+                @include col(10,12, $align:middle, $last:true);
             }
 
             .Person {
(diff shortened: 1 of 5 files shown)


src/components/panes/ActionReminderPane.scss
```diff
@@ -37,23 +37,25 @@
 
             .Avatar {
                 width: 5em;
-                height: auto;
+                height: 5em;
+                float: left;
```
jensborje: Here, we only need this:
`@include col(2,12, $align:middle, $first:true);`

src/components/panes/ActionReminderPane.scss
```diff
@@ -37,23 +37,25 @@
...
             }
 
-            .Person {
-                position: absolute;
-                top: 10px;
-                left: 60px;
+            .ActionReminderPane-remindedInfo {
+                float: left;
+                margin-left: 1em;
```
jensborje: Followed by this:
`@include col(10,12, $align:middle, $last:true);`

src/components/panes/ActionReminderPane.scss
```diff
@@ -37,23 +37,25 @@
...
+            }
 
+            .Person {
                 cursor: pointer;
-                font-size: 1.1em;
+                font-size: 1.4em;
                 font-weight: bold;
```
jensborje: In most cases like this, we use a larger font-size rather than bold. I suggest removing the `font-weight` and do `font-size: 1.7em`.

PR #498, opened 2017-03-02
Missing survey options

richardolsson:
This PR adds survey options that are in the API but were previously not accessible through the Organize UI:

* Whether to allow anonymous signatures on surveys
* Widget settings (`response_config`) for questions

### DIFF
--- a/src/components/forms/SurveyForm.jsx
+++ b/src/components/forms/SurveyForm.jsx
@@ -18,10 +18,17 @@ export default class SurveyForm extends React.Component {
 
     render() {
         let survey = this.props.survey || {};
-        let options = {
+        let accessOptions = {
             'open': 'forms.survey.accessOptions.open',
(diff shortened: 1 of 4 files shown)


src/components/forms/SurveyForm.jsx
```diff
@@ -31,16 +38,31 @@ export default class SurveyForm extends React.Component {
...
         );
     }
 
     getValues() {
-        return this.refs.form.getValues();
+        let values = this.refs.form.getValues();
+        values.allow_anonymous = (values.signature == 'anon');
+        delete values['signature'];
+
+        console.log(values);
```
jensborje: Logging needs to be removed.

richardolsson:
Good catch! Very sloppy of me. If you're happy with the changes, please approve and I will go ahead and merge. :)

Issue #528, opened 2017-03-20
Can't create survey question


richardolsson:
An exception is thrown when trying to create a new question.

![image](https://cloud.githubusercontent.com/assets/550212/24090773/8f27bc80-0d43-11e7-8ee8-bef16853370d.png)

Issue #524, opened 2017-03-19
Can't drop on empty actions


richardolsson:
The drop zone in `ActionListItem`doesn't seem to exist (or be infinitesimal) when there are no participants. This means new participants can't be dropped onto the action. This is most likely a result of the recent changes to the drop zone by @jensborje.

Issue #529, opened 2017-03-20
ActionListItem not showing all participants


richardolsson:
Because the max-height is static at `12em` in it's expanded state, the participant list in `ActionListItem` can never display more than 12 (4x3) participants. See screenshot of action with 21 participants where only the first 12 are shown.

![image](https://cloud.githubusercontent.com/assets/550212/24091156/8caddf18-0d46-11e7-8c58-dd818eed6fe1.png)

Is this by design, @jensborje?

jensborje:
I went with 12em more as a "large height" since we need a target height for animation. I see no problem with increasing it a bit. Too large and it will affect the animation duration though.

For the rare cases when the participant amount is exceptionally high, like on Mayday, I do think the action pane (which opens on click) will be a more relevant view for managing people. With that said, we perhaps don't need to be able to show all always? A label "click to view all" could also work.

PR #530, opened 2017-03-20
Misc bugs

richardolsson:
This PR fixes two critical bugs, #524 and #528, ahead of deploying a new version mid-sprint.

UPDATE: Also fixes undocumented bug in `updateOrAddListItems()` utility function.


richardolsson:
Merging without review because @jensborje is away, but would still like to discuss the changes to `ActionListItem` at some point. See separate issue #529.

PR #557, opened 2017-04-07
Survey query filters

richardolsson:
This PR adds UI "Smart search" query filters related surveys.


richardolsson:
Changes submitted, including styling specific for this particular use of `TextInput`. However, in the future the input components should be refactored to have their own styling, as tracked by #568.

PR #604, opened 2017-05-19
Search improvements

richardolsson:
This PR includes several fixes and improvements to the real-time search in Organize. Searching is now fuzzy, limited to 20 matches, and the groundwork has been laid for displaying the matching key in the search results in a future interface update.


richardolsson:
Merging without review due to reviewer unavailability.

PR #642, opened 2017-10-02
Canvass: mock createRoute

richardolsson:
This PR adds a mock `createRoute()` action (no API connection) which creates a route from a set of addresses. It also updates the way IDs are generated for (dummy) routes to more closely mimic the new alphanumerical IDs from the canvass API.


richardolsson:
This was an accidental pull request (wrong base). The correct one is #643.

PR #659, opened 2017-10-13
Use InfoList on more places

### LINKED ISSUE #627: General purpose PaneInfoList
Create a general-purpose info-list component to use in panes. Markup and styling for this type of content is currently repeated individually in every pane that uses it, e.g. `PersonPane` and `RoutePane`.

kesla:
Fixes #627

### DIFF
--- a/src/components/misc/InfoList.jsx
+++ b/src/components/misc/InfoList.jsx
@@ -33,8 +33,7 @@ export default class InfoList extends React.Component {
                     }
 
                     return (
-                        <li className={ className } key={item.name}
-                            onClick={ item.onClick }>
+                        <li className={ className } key={item.name}>
                             {value}
                         </li>
                     );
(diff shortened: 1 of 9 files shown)


src/components/misc/InfoList.jsx
```diff
@@ -16,10 +17,17 @@ export default class InfoList extends React.Component {
                     let className = 'InfoListItem InfoListItem-' + item.name;
                     let value = item.value;
 
-                    if (item.msgId) {
+                    // use custom value if set, using the msgId as a fallback
+                    if (!value && item.msgId) {
                         value = this.props.intl.formatMessage({ id: item.msgId }, item.msgValues);
                     }
 
+                    if (value && (item.onClick || item.href)) {
```
richardolsson: Great solution! Now that you've implemented this, please remove `onClick` on line 37. We discovered yesterday that `item.onClick` was ignored, and I added it to line 37 as a quick fix.

kesla: Thanks, done!

PR #660, opened 2017-10-13
Improve performance when dispatching a lot of events

### LINKED ISSUE #543: Select all in list is slow
This is because all items are selected individually. This would be greatly improved if an action was added to bulk select, i.e. add/remove multiple items simultaneously to a selection in the store with a single action.

kesla:
By default, the `.subscribe()` callback is run for each dispatch.

This PR changes the behaviour to call the callback at most once in a tick, meaning that if multiple actions are dispatched synchronously we see a large performance increase.

This means that selecting/deselecting all items in a list is a lot faster (see #543) without changing how we dispatch events.

See https://github.com/tappleby/redux-batched-subscribe is the redux enhancer used for this, linked to from http://redux.js.org/docs/faq/Performance.html#performance-update-events.

Fixes #543


richardolsson:
Interesting!

Correct me if I'm wrong, but with this approach, even if React is not re-rendered after each select action, the store logic is still far from optimized?

My idea of how to optimize this was to modify the `addToSelection()` action to use an array instead, and to `concat()` that array in the store, meaning only one action, one store mutation and one re-render would be required.

Is it your assessment that this is unnecessary? Is the repeated actions and store mutations not a problem, now that the repeated re-render has been optimized away?

kesla:
@richardolsson yeah, I think it's unnecessary. There's logic being run inside of react-redux on each subscribe, making it slow. But updating the state and running the logic within the redux reducer is really quick.

Another good thing here is that this is super flexible - we can in the future change the store in other ways, from multiple actions, and know that we will only update the UI once. I think that's a win from a maintenance POV.

kesla:
But tbh I chose this approach here since it was easiest and simplest to implement, making the smallest diff in the code, but at the same time fixing the underlying issue.

richardolsson:
Sounds like a great solution! Did you manage to create a case where you could measure an improvement in performance?

kesla:
@richardolsson Yup,

I expanded the # of items in the people list view (`/people/list`) and selecting/deselecting all appears instant now, but had a noticeable delay previous

PR #663, opened 2017-10-18
Canvass visit log

richardolsson:
This PR implements the visit log which lets the user follow progress during canvassing.


richardolsson:
This PR has become really big and diverse. Merging and will start new pull requests for more well-defined improvements.

Issue #730, opened 2018-02-26
Organization name on dashboard


richardolsson:
Add the name of the organization big on the Organize Dashboard. One user did not realize she was working in organization X, when she thought she was organizing Organization Y.

richardolsson:
Claiming this to let you focus on organization pages, @Bagera.

Issue #658, opened 2017-10-11
Playback centered to Malmö


jensborje:
Even though a campaign has planned actions on another city, it stills centers to Malmö.

Example: Most actions are planned in Malmö, but two is also in Lund. However, this is not shown since the map is centered to Malmö.

richardolsson:
I'm claiming this to let you focus on the organization pages, @Bagera.

PR #744, opened 2018-04-21
Misc improvements

richardolsson:
This PR contains a number of smaller improvements to the Organize interface.

### Manage section
The manage section that was implemented in the last release but hidden, is now made visible.

### Welcome message on dashboard
As outlined in #730, a message is added on the dashboard that includes the name of the current organization.

image

### Campaign playback
Fixes three issues related to the `CampaignPlayer`. One undocumented issue related to how the order in which actions were stored affected the playback span, one missing import in `CampaignPlaybackPane` and a third one as described in #658.

### Alert messages when unauthorized
Adds a framework for alert messages to display in the bottom left corner, and logic to display a message whenever an API request results in a 403 (unauthorized) error.

image

Issues #745, #746, #747 and #748 describe future improvements to and use cases for this framework.

### DIFF
--- a/locale/alerts/en.yaml
+++ b/locale/alerts/en.yaml
@@ -1,0 +1,5 @@
+error:
+    rejected:
+        http403: |-
+            You attempted something you're not authorized to do. Please verify
+            your access level or speak to a more senior official.
--- a/locale/alerts/sv.yaml
+++ b/locale/alerts/sv.yaml
@@ -1,0 +1,5 @@
+error:
+    rejected:
+        http403: |-
+            Du försökte göra något som inte är tillåtet. Var god kontrollera din
+            behörighetsnivå eller tala med någon ansvarig.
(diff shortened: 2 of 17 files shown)


Bagera:
Looks good.
The only things I can find that could be a bit different are:
- AlertMessages-messageItem-inOutAnimation: The keyframes have some duplication and should probably use transform instead of setting different `top` to get smoother animation.
- The recalc of map bounds could be a function to avoid repeat

richardolsson:
Great feedback, @Bagera! I have optimized the CSS animation now.

The map bounds recalculation should not be identical in two places, but use different props, which it didn't. Your comment alerted me to this, and I have fixed it now. It could still be broken out into a function, but I don't think it's necessary for such a tiny repetition.

In the process I came across another bug and fixed that too.

Please review again and approve, if you think it's ok. :)

richardolsson:
Thank you! Merging. :)

PR #759, opened 2018-05-08
Enhancement/filters

Bagera:
Still working in here but would want input.
The change is very subtle.
Maybe we want to show what filters are active? As tags besides the button?

Show filter when no filter applied (no change):
image

Show filters when filter is applied:
image

### DIFF
--- a/locale/panes/en.yaml
+++ b/locale/panes/en.yaml
@@ -1,3 +1,4 @@
 filterApplyButton: Apply filter
+filterButtonChange: Change filters
 filterButtonHide: Hide filters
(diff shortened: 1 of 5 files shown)


Bagera:
Blue button:
image

@richardolsson  I can't  find that bug. It reverts to no filter as far as I can see.

richardolsson:
We only wanted to change the color of the icon, not the whole button. But I think this is fine. What do you think, @jensborje?

Bagera:
New version with colored icon:
image

@jensborje  Like this?

PR #764, opened 2018-05-24
Enhancement/merge people

### LINKED ISSUE #756: PersonPane after merging
Open the master person object in `PersonPane` after successfully merging people.

Bagera:
fixes #756

### DIFF
--- a/src/store/view.js
+++ b/src/store/view.js
@@ -253,8 +253,14 @@ export default function viewState(state = null, action) {
     }
     else if (action.type == types.MERGE_PERSON_DUPLICATES + '_FULFILLED') {
         // Close the relevant pane
+        let openPanes = state.panes.filter(paneData => paneData.id != action.meta.paneId);
+        // Open the "master" person pane
         return Object.assign({}, state, {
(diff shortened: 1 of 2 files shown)


src/store/view.js
```diff
@@ -253,8 +253,14 @@ export default function viewState(state = null, action) {
     }
     else if (action.type == types.MERGE_PERSON_DUPLICATES + '_FULFILLED') {
         // Close the relevant pane
+        let openPanes = state.panes.filter(paneData => paneData.id != action.meta.paneId);
```
richardolsson: Instead of first closing and then opening, it's better to `map()` all panes, ignoring those that are irrelevant to the operation and replacing the closing one with the new one. Why? Because that way, it will always be opened in the right place. The way it's implemented now in this patch, the new pane will always be the last one, i.e. opened furthest to the right, even if the merge pane is not the rightmost one.

richardolsson:
Actually, merging this even though it's not perfect. I want to get it into the release, and we can always improve it later.

PR #787, opened 2018-06-08
Fix heightproblem in FireFox

### LINKED ISSUE #780: Import rows cut off in Firefox
image

Bagera:
Fixes #780


richardolsson:
Is this still working as expected in other browsers after this fix?

PR #813, opened 2018-06-29
Search via API

### LINKED ISSUE #664: Fuzzy search only as fallback
Some are surprised by the fuzzy search. Perhaps only use it when no "real" results are returned? Might also mean a performance improvement.

### LINKED ISSUE #726: Search is slow
A user reports that search is really slow. They're right. This could potentially be fixed by #664, but even after that we should probably profile for performance issues.

### LINKED ISSUE #743: Search for queries first
When searching for people, search for Smart Search query objects before searching for person objects. Just change the order in the search queue. Oftentimes, queries are fewer in count, so if a user searches for queries waiting through the people will take a lot of time, but for someone searching for people searching through the queries won't be a big hit.

richardolsson:
This PR replaces the old websocket searching implementation with a new set of search endpoints that are being added to the Zetkin Platform API.

Closes #743  
Closes #664
Fixes #726


richardolsson:
Implementation progress by data type:

- [ ] Action day (should happen locally)
- [ ] `AssignedRoute`
- [ ] `CallAssignment`
- [ ] `Campaign`
- [ ] `CanvassRoute`
- [ ] `CanvassAssignment`
- [ ] `Location`
- [x] `Person`
- [x] `PersonQuery`
- [ ] `Survey`
- [x] `SurveySubmission`

richardolsson:
Holding off on the Canvass-related search procs for now. Most of canvass searching is based on alphanumeric IDs which are not searchable at this point.

Issue #812, opened 2018-06-28
Exception when importing actions


richardolsson:
Probably due to some cells containing numbers instead of strings.

```
Uncaught (in promise) TypeError: str.split is not a function
    at parseTime (main.js:42)
    at main.js:42
    at Array.map (<anonymous>)
    at processDataRows (main.js:42)
    at actionImport (main.js:42)
    at main.js:59
    at computeNextEntry (<anonymous>:1:34000)
    at recomputeStates (<anonymous>:1:34300)
    at <anonymous>:1:37922
    at Object.dispatch (main.js:31)
```

A simple `String(str)` or `str.toString()` (may fail) should suffice. Also, errors should be caught and reported in the interface instead of crash.

PR #819, opened 2018-07-03
action importing: Cast parseTime input to String before using it as a String

olleolleolle:
This PR does one part of #812: Cast possible string input in `parseTime(str)` to String in order to try splitting on a regex.

### DIFF
--- a/src/store/actionImport.js
+++ b/src/store/actionImport.js
@@ -311,7 +311,7 @@ function parseTime(str) {
         return null;
     }
 
-    const fields = str.split(/[\.:]+/);
+    const fields = String(str).split(/[\.:]+/);
     if (fields.length && fields.length <= 3) {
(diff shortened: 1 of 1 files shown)


richardolsson:
Thanks for suggesting a solution to this! I'm sorry I didn't see this until now.

I fixed this another way, i.e. by using the plaintext value from the XLSX parsing. This was necessary because when we use the numeric, "raw" value, it differs vastly from the formatted value that users expect, and casting it to string is not enough.

For example, a date from Google spreadsheets that looks like "2018-07-07" will be numerically interpreted as `43288` (don't ask my why) and of course, casting that to `"43288"` is still not enough to parse it as a date.

See 2dc4cd8327cb74c338b669ca2e98dbfd0a49915e for my fix.

PR #826, opened 2018-07-07
Lazy-loaded lists

richardolsson:
This PR implements lazy-loading support in lists, and uses that feature to lazy-load action participants and responses in `ActionList`.


richardolsson:
Tested with real-life data. Seems to be ready for merge.

PR #834, opened 2018-07-21
Whitespaces during people merge

### LINKED ISSUE #799: Ignore surrounding whitespace in merge
The `MergePeoplePane` treats two values that are identical except for surrounding whitespace as different. This means that the user is asked to pick one of the two options, although they look identical. Steps to reproduce:

1. Log into Organize using dummy data, and navigate to People section
2. Create a new person with first name "Jane " (note the extra space after Jane), last name Doe, and e-mail testuser@example.com. This is a duplicate
3. Navigate to the Manage sub-section and the Duplicates tab. Start a search for duplicates.
4. Select to merge the two (or more?) Jane Doe objects

Expected results: First name is definitely Jane, so it should be pre-defined.
Actual results: You are asked to select one of two first names, "Jane" or "Jane ", which both appear exactly the same in the `<select>` that renders them.

Fix this by removing leading and trailing whitespace (e.g. using regex) before comparing them in `MergePeoplePane`.

### LINKED ISSUE #800: Ignore surrounding whitespace in duplicate search
Two person objects are not deemed similar enough when named the same, if one has an extra leading or trailing space in their name. Steps to reproduce:

1. Log into Organize using dummy data, and navigate to People section
2. Create a new person with first name "Jane " (note the extra space after Jane) and last name Doe. Add no other information. Identical first and last names should be enough for the algorithm to consider them a duplicate.
3. Navigate to the Manage sub-section and the Duplicates tab. Start a search for duplicates.

Expected results: Jane Doe should appear as a likely duplicate, including both the original object and the one created in step 2.
Actual results: The two objects are not considered duplicate because only the last names are deemed identical.

Fix this by removing leading/trailing whitespace before comparing in the duplicate finding API. Consider doing this in a pre-pass, to avoid having to do it more than once for every object as they're iterated.

fantratt:
Fixes issues regarding surrounding whitespace in people merges. 

Fixes #799  
Fixes #800

### DIFF
--- a/src/components/panes/MergePeoplePane.jsx
+++ b/src/components/panes/MergePeoplePane.jsx
@@ -97,6 +97,7 @@ export default class MergePeoplePane extends PaneBase {
                 let values = objects
                     .map(o => o[field])
                     .filter(val => !!val)
+                    .map(val => val.trim())
                     .filter((val, idx, arr) => arr.lastIndexOf(val) === idx);
 
                 let valueElem = null;
--- a/src/server/api/duplicates.js
+++ b/src/server/api/duplicates.js
@@ -15,8 +15,8 @@ const dupApi = express();
 const calcSimilarity = (master, other) => {
     let score = 0;
 
-    if (master.first_name == other.first_name) score += 6;
-    if (master.last_name == other.last_name) score += 6;
+    if (master.first_name.trim() == other.first_name.trim()) score += 6;
+    if (master.last_name.trim() == other.last_name.trim()) score += 6;
     if (master.ext_id && master.ext_id == other.ext_id) score += 20;
 
     // Add 6 for identical e-mails, subtract 2 for different


src/components/panes/MergePeoplePane.jsx
```diff
@@ -97,6 +97,7 @@ export default class MergePeoplePane extends PaneBase {
                 let values = objects
                     .map(o => o[field])
                     .filter(val => !!val)
+                    .map(val => val.trim())
```
richardolsson: Nice solution!

src/server/api/duplicates.js
```diff
@@ -15,8 +15,8 @@ const dupApi = express();
 const calcSimilarity = (master, other) => {
     let score = 0;
 
-    if (master.first_name == other.first_name) score += 6;
-    if (master.last_name == other.last_name) score += 6;
+    if (master.first_name.trim() == other.first_name.trim()) score += 6;
```
richardolsson: This is good. In the future, we might want to optimize this by not repeating it for every iteration. But for the time being we don't really have any performance issues motivating such optimizations.

PR #859, opened 2018-08-05
PersonList connects onLoadPage implicitly - rely on that

olleolleolle:
This commit only removes the explicit pointing out which onLoadPage callback would be connected. It is already done, in the component.


richardolsson:
Closing as per afk discussion. :)

PR #867, opened 2018-08-10
Optimized duplicate search

### LINKED ISSUE #866: Optimize duplicate search
The algorithm to search for duplicates is too slow for organizations with more than 1000 or so people in them. Try to find a way to optimize it, or bucket it so that it stops after a certain time and returns whatever it may have found, even if it's not complete.

richardolsson:
This PR optimizes the duplicate search using the following strategies:

* Calculate normalized values for person attributes once, and reuse
* Don't pop/splice lists, use indices and flags instead
* Don't create a duplicate object unless duplicates are found

## Benchmarks
These benchmarks were measured using real-life data in a mid-size and large organization respectively. The times in seconds are measured using the Chrome dev tools when making a request to the duplicate search endpoint for each of the organizations before and after the change.

| Case | Trial 1 | Trial 2 | Trial 3 | Trial 4 | Trial 5| Avg boost
|:-----|-------:|-------:|-------:|-------:|-------:|--------:|
| Mid-size org, before | 9.23 | 8.15 | 4.65 | 4.75 | 4.66 |
| Mid-size org, after | 1.77 | 1.63 | 1.67 | 1.59 | 1.62 | 4x
| Large org, before | 35.11 | 14.16 | 15.28 | 13.63 | 15.59 |
| Large org, after | 1.75 | 1.53 | 1.52 | 1.46 | 1.71 | 12x

Closes #866

### DIFF
--- a/src/server/api/duplicates.js
+++ b/src/server/api/duplicates.js
@@ -42,6 +42,7 @@ dupApi.get('/:orgId/people', (req, res, next) => {
             let duplicates = [];
 
             let people = result.data.data.map(p => {
+                p.consumed = false;
                 p.n_phone = p.phone? p.phone.replace(/\D/g, '') : null;
                 p.n_email = p.email? p.email.toLowerCase() : null;
                 p.n_first_name = p.first_name.trim().toLowerCase();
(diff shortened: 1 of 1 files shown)


richardolsson:
Thank you @olleolleolle! Will go ahead and merge. :)

Issue #875, opened 2018-08-14
Cap action export


richardolsson:
Exporting actions from the action list requires a lot of data to be retrieved behind the scenes. The operation is likely to time out for huge amounts of actions. Cap this at 100 or so actions.

The best way to achieve this is to check in `BulkOpsPane` whether the count is > 100 (for the relevant bulk op) and disable the execute button.

This could be done using some sort of nice framework, or by just hardcoding.

PR #876, opened 2018-08-14
BulkOpPane: avoid action exports of more than 100 items

olleolleolle:
This PR introduces a hard-coded limit of "max 100 actions in an actions export".

- an explanation is rendered in the Footer of the pane
- a warning color (SCSS orange) visualizes the warning-ness

## Needs

- A clearer indication that this is unsupported.
- Not a `<Button>`.

Example (where the limit is set to 10 not 100, in order to hit the limit)

image

See #875.

### DIFF
--- a/locale/panes/bulkOp/en.yaml
+++ b/locale/panes/bulkOp/en.yaml
@@ -18,3 +18,4 @@ headers:
                 =1 {one person}
                 other {# persons}}
 submitButton: Perform operation
+tooLargeActionExport: Action Export too large, max number supported is { maxActionsSupportedInExport }
--- a/src/components/panes/BulkOpPane.scss
+++ b/src/components/panes/BulkOpPane.scss
@@ -1,0 +1,8 @@
+.PaneBase {
+    .PaneBase-footer {
+        button.tooLargeActionExport {
+            background-color: $c-ui-warning;
+            border-color: darken($c-ui-warning, 10);
+        }
+    }
+}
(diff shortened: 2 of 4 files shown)


locale/panes/bulkOp/en.yaml
```diff
@@ -18,3 +18,4 @@ headers:
                 =1 {one person}
                 other {# persons}}
 submitButton: Perform operation
+tooLargeActionExport: You can at most export { maxActionsSupportedInExport } Actions.
```
richardolsson: I think it would make sense to have this under `action.export` in the YAML structure, in which case it can also have a shorter key name, like `action.export.tooLarge`.

src/components/panes/BulkOpPane.scss
```diff
@@ -0,0 +1,8 @@
+.PaneBase {
```
richardolsson: This entire file should no longer be relevant?

src/components/bulk/config/ActionExportBulkOpConfig.scss
```diff
@@ -0,0 +1,16 @@
+.PaneBase {
```
richardolsson: There is no need to wrap the SCSS in this file in a `.PaneBase` selector. In fact, the file itself is just top-level CSS, meaning that if anything, it should be wrapped in an `.ActionExportBulkOpConfig` selector.

However, because the next-level select, `.ActionExportBulkOpConfig-maxActionsWarning` is already specific enough to select the correct element, there is no need for any wrapping.

Conclusion: Remove the wrapping `.PaneBase {}` and unindent the entire content.

PR #879, opened 2018-08-19
show loading indicator in all actions pane

### LINKED ISSUE #873: Loading indicator while loading actions
There is nothing on  `AllActionsPane` indicating that actions are being loaded. Display a loading indicator while the action list is pending.

fantratt:
fixes #873

### DIFF
--- a/src/components/sections/campaign/AllActionsPane.jsx
+++ b/src/components/sections/campaign/AllActionsPane.jsx
@@ -63,7 +63,7 @@ export default class AllActionsPane extends CampaignSectionPaneBase {
 
     renderPaneContent(data) {
         let actionList = this.props.filteredActionList;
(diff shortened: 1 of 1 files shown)


src/components/sections/campaign/AllActionsPane.jsx
```diff
@@ -82,8 +83,10 @@ export default class AllActionsPane extends CampaignSectionPaneBase {
         else if (actionList.items.length == 0) {
             endDate = startDate.clone().addDays(8 * 7);
         }
-
-        if (this.state.viewMode == 'cal') {
+        if (actionList.isPending) {
```
richardolsson: This will never happen, because on line 66, an `actionList` without `items` will prevent rendering altogether.

Issue #404, opened 2017-01-25
Implement full support for random filters


richardolsson:
The `random` query filter has support for seeding and fractional sizes, which are not yet supported in the Organize UI.

richardolsson:
From the query DSL documentation:

### Random (`random`)
The random filter selects a random subset of the people in an organization. The
size of the subset is defined in the config.

```json
{
    "type": "random",
    "config": {
        "size": 0.7,
        "seed": "abc123"
    }
}
```

The `size` property defines how many people to return and must be a positive
number. Integers above 1 are interpreted as a fixed size, meaning that a value
of 5 will result in at most 5 persons being returned. Fractions between 0 and 1
are interpreted as a percentage of the total number of people in the
organization, so a value of 0.5 will return 50% of the people.

The `seed` property can be used to refresh the random selection. The selection
is stable in the sense that given the same seed, the same subset will always be
returned, as long as the database doesn't change.

If the seed changes, the subset will be very different. If the database changes,
e.g. by adding or removing people, the random subset will at most change by the
number of people added and removed.

The default seed is the empty string.

richardolsson:
Suggested improvements to the current UI:

- [ ] Add a button to refresh the random selection (update the seed)
- [ ] Clarify choice between percentage/exact number (e.g. using a `<select>` to choose method)

richardolsson:
Fixed by #927

PR #927, opened 2018-11-12
Feature/random filters#404

alexander-js:
This PR adds the following to the Random Filter:

- The ability to swap between two units. `People`, `Percent`
- A button for randomizing the output via seed generation.
- An amount cap of 99 if the unit is set to `Percent`.
- Corresponding locale mappings. (In need of review)

Implements #404

### DIFF
--- a/src/components/filters/RandomFilter.jsx
+++ b/src/components/filters/RandomFilter.jsx
@@ -68,7 +68,6 @@ export default class RandomFilter extends FilterBase {
             ? parseFloat('0.' + amount)
             : amount;
 
-        console.log("config", { size, amount, seed, unit })
         return { size, amount, seed, unit };
     }
 
--- a/src/components/filters/RandomFilter.jsx
+++ b/src/components/filters/RandomFilter.jsx
@@ -12,46 +12,75 @@ export default class RandomFilter extends FilterBase {
     constructor(props) {
         super(props);
 
+        const { amount, seed, unit } = this.mapConfigToState(props.config);
         this.state = {
-            amount: props.config.amount || 20,
-            seed: props.config.seed || makeRandomString(6),
-            unit: props.config.unit || 'people'
+            amount: amount || 20,
+            seed: seed || makeRandomString(6),
+            unit: unit || 'people'
         };
     }
 
-    componentReceivedProps(nextProps) {
-        if (nextProps.config !== this.props.config) {
-            this.setState({
-                amount: nextProps.config.amount,
-                seed: nextProps.config.seed,
-                unit: nextProps.config.unit,
-            });
+    shouldComponentUpdate = (_, nextState) => nextState !== this.state;
+
+    componentWillReceiveProps({ config }) {
+        const { state, mapConfigToState } = this;
+        const mappedState = mapConfigToState(config);
+
+        // We can't simply compare the objects for diffs with === because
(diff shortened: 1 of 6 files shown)


src/components/filters/RandomFilter.jsx
```diff
@@ -2,47 +2,108 @@ import React from 'react';
...
         ];
     }
 
     getConfig() {
-        return {
-            size: this.state.size,
-            seed: this.state.seed,
-        };
+        const { amount, seed, unit } = this.state;
+        // We're doing this because percentages are interpeted
```
olleolleolle:
```suggestion
        // We're doing this because percentages are interpreted
```

src/components/filters/RandomFilter.jsx
```diff
@@ -2,47 +2,108 @@ import React from 'react';
...
+        // as fractions in the backend.
+        const size = unit === 'percentage'
+            ? parseFloat('0.' + amount)
+            : amount;
+
+        return { size, amount, seed, unit };
```
richardolsson: This should not return more than the properties allowed in the filter configuration that we send to the back-end, i.e. `seed` and `size`.

Instead of reyling on `amount` and `unit` being stored like this, derive them from the `size` in the constructor and `componentWillReceiveProps()`.

src/components/filters/RandomFilter.jsx
```diff
@@ -2,47 +2,108 @@ import React from 'react';
...
+                className="RandomFilter__size"
                 labelMsg="filters.random.size"
-                value={ this.state.size }
-                onValueChange={ this.onSizeChange.bind(this) }
-                />,
+                value={ amount }
+                onValueChange={ this.onSizeChange }
+            />,
+            <SelectInput key="unit" name="unit"
+                className="RandomFilter__unit"
```
richardolsson: Class name not according to code style.

src/components/filters/RandomFilter.jsx
```diff
@@ -2,47 +2,108 @@ import React from 'react';
...
+        const { unit, amount } = this.state;
+        const msg = id => this.props.intl.formatMessage({ id });
+        const units = {
+            'people': msg('filters.random.units.people'),
+            'percentage': msg('filters.random.units.percentage'),
+        };
+
         return [
             <IntInput key="size" name="size"
+                className="RandomFilter__size"
```
richardolsson: Class name not according to code style.

src/components/filters/RandomFilter.jsx
```diff
@@ -2,47 +2,108 @@ import React from 'react';
...
+                onValueChange={ this.onSizeChange }
+            />,
+            <SelectInput key="unit" name="unit"
+                className="RandomFilter__unit"
+                labelMsg="filters.random.units.label"
+                options={ units }
+                value={ this.state.unit }
+                onValueChange={ this.onUnitChange }
+            />,
+            <div className="RandomFilter__btnWrapper">
```
richardolsson: Class name not according to code style.

src/components/filters/RandomFilter.scss
```diff
@@ -0,0 +1,19 @@
+.RandomFilter {
+    &__size {
```
richardolsson: For suffixes like these, we usually use a single dash (e.g. `-size`).

PR #929, opened 2018-11-16
Feature/mobile interface

hannis:
Responsive styling for the header is now up and running. Please have a look!

### DIFF
--- a/src/components/header/search/Search.scss
+++ b/src/components/header/search/Search.scss
@@ -43,10 +43,32 @@
         &:focus {
             text-align: left;
         }
+
+        @include mini-screen {
+
+            // WebKit browsers
+            &::-webkit-input-placeholder {
+             color:  #fff;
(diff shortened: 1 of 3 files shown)


src/components/header/search/Search.scss
```diff
@@ -43,10 +43,32 @@
         &:focus {
             text-align: left;
         }
-    }
 
+        @include mini-screen {
+
+            // WebKit browsers
+            &::-webkit-input-placeholder {
+             color:  #fff;
```
jensborje: This sets a white text color to a white background. Is there a reason why? :)

hannis: This is now changed to transparent instead of white.

richardolsson:
Is this ready for merge, @jensborje? Please let me know tomorrow and I will push the button. 🙌

PR #930, opened 2018-11-17
Button for removing filter queries

### LINKED ISSUE #482: Delete person query
There is no way to delete a stored person query.

alexander-js:
This PR adds a button for removing filter queries in the EditQueryPane. 

Note that - while this **works** - it should eventually have proper error handling and a loading indicator. Alternatively, the async operation could be handled unbeknownst to the user in the background in a non-blocking fashion.

Implements #482


src/actions/query.js
```diff
@@ -73,6 +73,21 @@ export function updateQuery(id, data) {
     };
 }
 
+export function removeQuery(id) {
+    return ({ dispatch, getState, z }) => {
+        const orgId = getState().org.activeId;
+        const where = `orgs/${orgId}/people/queries/${id}`.split('/');
```
richardolsson: No need to `split()` and `...spread`. You can just pass the string directly to `z.resource()`. See README for the Zetkin SDK here: https://github.com/zetkin/zetkin.js

src/components/panes/EditQueryPane.jsx
```diff
@@ -43,6 +45,10 @@ export default class EditQueryPane extends PaneBase {
     }
 
     componentWillReceiveProps(nextProps) {
+        if (!nextProps.queryItem) {
+            return nextProps.onClose();
```
richardolsson: Use `this.closePane()` instead, for consistency with other panes.

PR #936, opened 2018-11-28
Feature/mobile interface

hannis:
Please have a look at these responsive changes. The section nav is a bit changed and I've also made som changes on the different lists on smaller screens. I also made some changes in RootPaneBase.scss to make it responsive.


hannis:
I've looked over alls things above and they are all fixed in my latest commit.

PR #947, opened 2018-12-13
Move styling to Input

### LINKED ISSUE #568: Input styling
The input components should have their own styling, not just inherit from `Form` et c.

jensborje:
This PR moves styling relating to input fields from a global mixin called `form-base` to be more component specific. This styling is now in the component `InputBase`. All Input components inherits this.

We are however still to use the global mixin `input-base`, since this is shared with components outside of this scope.

Also removed unnecessary use of `input-base` in some of the input components.

Fix #568

### DIFF
--- a/src/components/forms/inputs/SelectInput.jsx
+++ b/src/components/forms/inputs/SelectInput.jsx
@@ -31,10 +31,12 @@ export default class SelectInput extends InputBase {
         }
 
         return (
-            <select value={ this.props.value || '0' }
-                onChange={ this.onChange.bind(this) }>
-                { optionElements }
-            </select>
+            <div className="SelectInput">
+                <select value={ this.props.value || '0' }
+                    onChange={ this.onChange.bind(this) }>
+                    { optionElements }
+                </select>
+            </div>
         );
     }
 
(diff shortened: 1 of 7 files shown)


src/components/forms/inputs/SelectInput.scss
```diff
@@ -1,10 +1,9 @@
-select{
-    @include input-base;
+select {
```
richardolsson: Is it really a good idea to globally style `select` (and not `.SelectInput`)?

jensborje: No it isn't. All I did in this commit was to add a space, but we should make improvements when we notice them. Well spotted!

olleolleolle: 🦅 👁

src/components/forms/inputs/SelectInput.jsx
```diff
@@ -31,10 +31,12 @@ export default class SelectInput extends InputBase {
         }
 
         return (
-            <select value={ this.props.value || '0' }
-                onChange={ this.onChange.bind(this) }>
-                { optionElements }
-            </select>
+            <div className="SelectInput">
```
olleolleolle: React-beginner: Just learned about the  `className` property!

https://reactjs.org/docs/faq-styling.html

jensborje:
Updated the PR description. Hope it makes more sense now!

PR #950, opened 2018-12-21
Mobile Interface Dialog Styling

hannis:
This PR adds responsive styling to Dialog - Assignments and Call log and also Canvass - Assignments and Visit Log.

+ I also added display none to the span with text (old actions) in Actions so that it don't show on smaller screens.


richardolsson:
I'm reviewing this now so you don't have to @jensborje. But do let me know if you also want to and I can hold off the (expected) merge. 😄

Issue #898, opened 2018-08-26
Responsive/mobile layout


richardolsson:
Finalize, improve and release the mobile layout. Work on this started early in the process of developing Organize, but was never finished.

jensborje:
## UX analysis ##
Note: Not all in the following is to be considered within the scope of this issue.

### What differs from desktop usage? ###
Desktop usage is in a workspace, a controlled and (hopefully) focused environment. The tool has the users primary attention. This makes is suitable for planning and analyzing.

Mobile usage is in a more uncontrolled and dynamic environment. It can be on a bus, on the street or in a sofa next to the workspace. Often, dialogs with others are close at hand. The attention of the user could therefore be potentially on another person primarily or shared with other needs such as orientation. This gives the mobile usage a high potential in situations where a user needs status reports as support to a conversation, or to carry through the result of a conversation or a outer situation immediately, such as adding a person to a task you agreed to.

Since the environment is potentially changing by outer circumstances, usage duration tends to be short. This makes mobile usage not suitable for activities such as planning or analyzing.
The potentially secondary attention, and the limited usage duration, highlights the needs for accessibility for tasks that is often used in situations related to the usage environment. This increases the need of a task driven rather than a conceptual driven navigation which is more often used when designing for desktop usage.

These different needs between device usage clashes however with another need - to build an understanding of the tools features and recognition across the platform. This usually results in attempts to keep the experiences as similar as possible. A potential reason for this could be that it also aligns with business needs such as marketing. A compromise often seen is device specific shortcuts.

### Not yet investigated ###
- Screen limitations VS handheld potential in terms of intuitive navigation

### Potential needs within the mobile usage scope ###

* Ongoing projects
  * Monitor progress
  * Manage participants


* Mobilize
  * Actions
    * Edit, add and remove actions
    * Manage participants
    * Send reminders


* Member management
  * Look up members
    * Smart searches
    * Update member info
  * Manage groups
    * Edit, add and remove groups
    * Update group members
  * Review survey submissions

jensborje:
### Pages relevant for mobile usage ###

**/people**
/list
~~/import~~
/groups
/manage

**/campaign**
/actions
~~/distribution~~
~~/playback~~
/campaigns

**/dialog**
/assignments
/calls

**~~/maps~~**
~~/overview~~
~~/locations~~

**~~/survey~~**  *
~~/surveys~~
~~/submissions~~

**/canvass**
/assignments
~~/routes~~
[...]

Issue #938, opened 2018-11-30
info.png is missing


stefanfisk:
Referenced here: https://github.com/zetkin/organize.zetk.in/blob/b4abd85c58d2b79968449ae76307ca6a6f736dec/src/components/lists/items/ActionListItem.scss#L200

richardolsson:
Do you know what this is about, @jensborje? I'm not even sure what that is. :)

Maybe this is something that @hannis can look at and fix while going over the `ActionList` for the mobile layout?

jensborje:
Seems to be outdated code that I've forgot to remove when i made a refactor a while back. Originates from Victors initial work 3 years ago.

hannis:
The outdated code has now been removed in my latest commit in the feature/mobile-interface.

PR #952, opened 2019-01-12
Feature/mobile interface

hannis:
Hello,

This PR is all about the responsive content in the Panes that are going to be used for mobile usage. I've made mostly minor and detailed responsive styling changes to make them look good on small screens. 

@jensborje and me went through the entire list of Panes to make sure that we haven't forgotten any.

Please not that the JS-stuff that will make the Panes land on the correct spot are not done yet. @jensborje and me are going to start working on those issues next week.

PR #951, opened 2018-12-28
Action calendar styling 

hannis:
I've made some changes in the Action Calendar styling. It had turned the other way round due to new media queries (small screens had the desktop styling and vice versa). I hope and think it looks as it should now, but please take a closer look. 

actions_calendar

PR #956, opened 2019-01-28
Mobile interface

jensborje:
This PR builds upon work in the same branch made by @hannis. It's ready to be merged into master, and comprises changes made over several now merged pull requests onto the `feature/mobile-interface` branch: #928, #929, #936, #939, #950, #951, #952, #954 and #955.

The main feature is to add mobile support. The app is still mainly made for desktop use but does now also allow mobile use for selected parts of the app that is relevant for non-desktop situations.

Besides a large amount of small changes in many panes and components in terms of margins and sizing, we have a couple of larger efforts.

* Canged the mixins for media queries from mobile first (though rarely used) to be desktop fist. Therefor, we also need to update several non-mobile parts in order to resolve issues caused by the switch.

* Moved more styling code to be more general in base components rather than in each pane.

* Refactored/simplified the pane management for mobile to only stack on top of each other.

* Adjusted complex list items to work on mobile.

* Mobile styling for search results

* Horizontal navigation bar for mobile with scrolling if needed

* Restyled close link for panes that works for both desktop and mobile.

* Removed hover dependency in order to see icons such as edit or remove. (see callers in /dialog/assignments/callassignment:4)

* ViewSwitch in a RootPane toolbar is restyled for mobile to become a sticky foooter to change it's view mode (see /people/manage)

* Set view mode in /campaign/actions to be list when on mobile and hide the View switch.

* Hide parts of the app that should not be used on mobile.

UX analysis: https://github.com/zetkin/organize.zetk.in/issues/898
Sketch: https://sketch.cloud/s/gAzqa/p/mobile

screen shot 2019-01-30 at 15 08 22

screen shot 2019-01-30 at 15 09 18

screen shot 2019-01-30 at 15 11 43

screen shot 2019-01-30 at 15 11 58

screen shot 2019-01-30 at 15 12 36

screen shot 2019-01-30 at 15 21 41

screen shot 2019-01-30 at 15 25 19


richardolsson:
Closes #898
Closes #938

Issue #970, opened 2019-02-20
Use search for RelSelectInput


richardolsson:
Instead of retreiving all objects of a certain type, use the search API for `RelSelectInput` (and `PersonSelectWidget`). This will require the search store to be refactored to be able to handle multiple searches, but would mean a huge performance improvement.

richardolsson:
Since #977, `PersonSelectWidget` now uses the search API, but other occurrences of `RelSelectInput` do not. This is fine for now because they are less urgent (there can be 100 000 people, but rarely 100 000 locations, campaigns, activities or other object types for which `RelSelectInput` is used). Postponing further development to a future release.

PR #977, opened 2019-02-22
Refactor search store to handle multiple searches

martinajsongren:
Refactor search store to handle multiple searches. 
Refactor PersonSelectWidget to use search API to retrieve people.

Related to #970

### DIFF
--- a/src/actions/search/index.js
+++ b/src/actions/search/index.js
@@ -5,10 +5,10 @@ import { SearchQueue } from './utils';
 
 let queue;
 
-export function search(query) {
+export function search(query, field) {
     return ({ dispatch, getState, z }) => {
         const orgId = getState().org.activeId;
-        const scope = getState().search.scope;
+        const scope = getState().search[field].scope;
         const lang = getState().intl.locale;
 
         if (queue) {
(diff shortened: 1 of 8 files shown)


src/components/header/search/Search.jsx
```diff
@@ -15,10 +15,10 @@ import {
     search,
 } from '../../../actions/search';
 
-
+const field = "top"
```
richardolsson: I would call it `FIELD_NAME` to clearly denote everywhere that it's a (global) constant.

src/actions/search/index.js
```diff
@@ -5,18 +5,18 @@ import { SearchQueue } from './utils';
 
 let queue;
 
-export function search(query) {
+export function search(query, field) {
```
richardolsson: Use the same argument order for all actions creators, i.e. `func(field, any, other, args)`.

richardolsson:
Great work! I will take a closer look at this tomorrow, but from just glancing over it quickly I can tell that some really great progress is being made. Good job today!

src/components/header/search/Search.jsx
richardolsson: Fixed by 9a470f15a85febd6ff8f9c776f3a040a8a2ca45c

src/actions/search/index.js
richardolsson: Fixed by 9a470f15a85febd6ff8f9c776f3a040a8a2ca45c

PR #981, opened 2019-03-03
Bulk create survey options

### LINKED ISSUE #922: Faster survey options
Add a way to bulk-create options for a question, using a textarea. For example, add a button called "add multiple" and show a textarea where every line will be one new option.

richardolsson:
This PR adds an interface to create multiple response options for a survey question at once. A new button to "add multiple" is added next to the existing "add response option" one. Clicking it presents a `textarea` where the user can type several options, one per line, before clicking a button to have them all created in one go.

Fixes #922

image

### DIFF
--- a/locale/misc/en.yaml
+++ b/locale/misc/en.yaml
@@ -49,8 +49,11 @@ surveyOutline:
             text: Text
     option:
         addButton: Add response option
+        addBulkButton: Add multiple
+        bulkInstructions: Type one option per line.
         cancelButton: Cancel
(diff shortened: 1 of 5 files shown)


richardolsson:
I'm particularly interested in hearing your views on the UX/visuals, @jensborje.

richardolsson:
That's a good suggestion! I will quickly implement that change and re-request review in a bit.

richardolsson:
It now looks like this: 

image

PR #983, opened 2019-03-03
Fix exception in call history filter

### LINKED ISSUE #914: Call history filter bug
The "between certain dates" seems to not work properly. Steps to reproduce:

1. Create a new smart search query
2. Add a Call history filter
3. Change the timeframe to be "between certain dates" (cited from memory)

Expected results: Two date fields should appear
Actual results: Nothing happens

richardolsson:
Fixes #914

### DIFF
--- a/src/components/filters/CallHistoryFilter.jsx
+++ b/src/components/filters/CallHistoryFilter.jsx
@@ -208,7 +208,7 @@ export default class CallHistoryFilter extends FilterBase {
                 break;
             case 'between':
                 after = today;
-                before = (new Date()).advance({ days: 30}).format('{yyyy}-{MM}-{dd}');
+                before = (new Date()).advance({ days: 30 }).format('{yyyy}-{MM}-{dd}');
                 break;
         }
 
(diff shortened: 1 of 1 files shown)


src/components/filters/CallHistoryFilter.jsx
```diff
@@ -208,7 +208,7 @@ export default class CallHistoryFilter extends FilterBase {
                 break;
             case 'between':
                 after = today;
-                before = (30).daysAfter(today).format('{yyyy}-{MM}-{dd}');
+                before = (new Date()).advance({ days: 30}).format('{yyyy}-{MM}-{dd}');
```
olleolleolle:
```suggestion
                before = (new Date()).advance({ days: 30 }).format('{yyyy}-{MM}-{dd}');
```

Balance the whitespace in { and }, in some way. I like "one space before and after content.

olleolleolle: Eh, `advance` - I don't know where that API came from, so here's something only using the Date object:

```js
let inOneMonth = new Date();
inOneMonth.setDate(inOneMonth.getDate() + 30);
```

olleolleolle: image

richardolsson:
The `Date.advance()` method is added by "Sugar.js" which we are using mainly for localized formatting.

PR #985, opened 2019-03-04
Fix bug in ActionList lazy loading

### LINKED ISSUE #969: Lazy loading bug when list changes
When a list changes, it does not trigger a re-load of the visible objects until they're scrolled out of view and back again, which may not be possible if the list is not long enough.

Steps to reproduce:

1. Navigate to the action list
2. Wait for participants to load into visible action items
3. Click to show past actions

Expected results: Past actions should show and participants for visible actions should load
Actual results: Past actions show, but participants do not load

richardolsson:
Fixes #969 

(See issue ticket for details)

### DIFF
--- a/src/components/lists/items/ActionListItem.jsx
+++ b/src/components/lists/items/ActionListItem.jsx
@@ -114,15 +114,13 @@ export default class ActionListItem extends React.Component {
     }
 
     loadSubData() {
-        const action = this.props.data;
-        const participants = this.props.participants;
-        const responses = this.props.responses;
+        const { data, participants, responses } = this.props;
 
         if (!participants) {
-            this.props.dispatch(retrieveActionParticipants(action.id));
+            this.props.dispatch(retrieveActionParticipants(data.id));
         }
         if (!responses) {
-            this.props.dispatch(retrieveActionResponses(action.id));
+            this.props.dispatch(retrieveActionResponses(data.id));
         }
     }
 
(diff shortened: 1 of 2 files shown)


richardolsson:
I'm taking a long shot that you'll be able to review this today, @olleolleolle, before I start preparing the release. Let me know if you can't. 😄

src/components/lists/items/ActionListItem.jsx
```diff
@@ -103,16 +103,26 @@ export default class ActionListItem extends React.Component {
...
+    componentDidMount() {
+        if (this.props.inView) {
+            this.loadSubData();
+        }
+    }
+
+    loadSubData() {
+        const action = this.props.data;
+        const participants = this.props.participants;
+        const responses = this.props.responses;
```
olleolleolle: Is destructuring assignment available here?

`const { action, participants, responses } = this.props;`

richardolsson: It should work, but it's something we haven't been using (consistently) before. I really think we need a general code style overhaul at some point seeing how javascript has evolved since this project started (almost exactly!) four years ago. That's the kind of work that I have planned for late 2019, although I guess it makes sense to start implementing it little by little already.

olleolleolle: (Thanks also for spotting the bug in my shoddy example!)

PR #1002, opened 2019-04-01
Survey submission filters

richardolsson:
This PR adds pagination and more filtering options for the survey submissions listing.

### DIFF
--- a/src/actions/surveySubmission.js
+++ b/src/actions/surveySubmission.js
@@ -17,7 +17,7 @@ export function retrieveSurveySubmissions(surveyId = null, page = 0, perPage = 2
 
         dispatch({
             type: types.RETRIEVE_SURVEY_SUBMISSIONS,
-            meta: { surveyId },
+            meta: { page, surveyId },
             payload: { promise },
         });
     };
--- a/src/actions/surveySubmission.js
+++ b/src/actions/surveySubmission.js
@@ -1,23 +1,29 @@
 import * as types from '.';
 
 
-export function retrieveSurveySubmissions(surveyId = null, page = 0, perPage = 20) {
+export function retrieveSurveySubmissions(surveyId = null, linked = null, page = 0, perPage = 20) {
     return ({ dispatch, getState, z }) => {
-        let orgId = getState().org.activeId;
+        const filters = [];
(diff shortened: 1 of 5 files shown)


src/actions/surveySubmission.js
```diff
@@ -1,22 +1,29 @@
...
-export function retrieveSurveySubmissions(surveyId = null) {
+export function retrieveSurveySubmissions(surveyId = null, linked = null, page = 0, perPage = 20) {
     return ({ dispatch, getState, z }) => {
-        let orgId = getState().org.activeId;
+        const filters = [];
+        const orgId = getState().org.activeId;
         let promise;
 
+        if (linked !== null) {
+            linked = linked? ((filters.linked == 'linked')? 1 : 0) : null;
```
olleolleolle: Style note: the positioning of the `?` operator in the ternary conditional tripped me up. "Is this some Babel precompile feature?"

src/store/surveySubmissions.js
```diff
@@ -8,50 +8,68 @@ import {
...
-                submissionList: createList(),
-            };
+        }
+    }
+    else if (action.type == types.RETRIEVE_SURVEY_SUBMISSIONS + '_FULFILLED') {
+        return Object.assign({}, state, {
+            submissionList: updateOrAddListItems(state.submissionList, action.payload.data.data, {
+                isPending: false,
+                error: null,
+                lastPage: Math.max(state.submissionList.lastPage, action.meta.page),
```
olleolleolle: This got me narrowing my eyes a little "What do we _know_ about these properties? How?" I don't have the answer, yet.

`Math.max` really does not like `undefined`:

```
Math.max(1,2,undefined)
NaN
```

`null` is OK, though:

```
Math.max(1,2,null)
2
```

src/actions/surveySubmission.js
richardolsson: Yeah, multi-level ternaries are not a great style choice. I'll reconsider in the future. :)

src/store/surveySubmissions.js
richardolsson: True! I think this entire framework for pagination needs to be revamped and this is something to consider when we do so.

PR #1005, opened 2019-04-14
Mass SMS crud

stefanfisk:
This covers CRUD for SMS distributions.

The credits UI still needs to be implemented, so once the 1100 dummy credits are gone sending will fail without there being any user visible explanations as to why.


src/components/panes/SmsDistributionCreditTransactionPane.jsx
```diff
@@ -0,0 +1,181 @@
...
+            description,
+
+            charge: {
+                amount: chargeAmount,
+                last4
+            } = {},
+
+        } = transactionItem.data;
+
+        const receiptUrl = 'http://example.com#not-implemented';
```
stefanfisk: I can't figure out how to implement this. The core endpoint is `/orgs/org_id:/sms_distribution_credit_transactions/:transaction_id/receipt`, which returns a PDF.

stefanfisk:
Push the credits UI stuff, now this should be complete as far as functionality goes, except for the receipt URLs which I've commented.

PR #1037, opened 2020-01-23
Render organizer message in CallListItem

### LINKED ISSUE #933: Render organizer message in CallList
When there is an `organizer_action_needed` flag on a call, it would be useful for the organizer to be able to see the `message_to_organizer` in the list, to more rapidly decide which tasks can be bulked together, which need to be prioritized, et c.

MikaelJohnAndersson:
This PR adds functionality for displaying the message to organizer in `CallListItem`.
Since `message_to_organizer` only is included in the data when retrieving a single call, `CallListItem` dispatches an action for retrieving additional data once the item is in view, displaying a loading (dots) animation while the request is pending.
In cases where there is no `message_to_organizer` data, a fallback message is rendered.

Closes #933

### DIFF
--- a/src/components/lists/items/CallListItem.jsx
+++ b/src/components/lists/items/CallListItem.jsx
@@ -1,15 +1,30 @@
+import { connect } from 'react-redux';
 import React from 'react';
 import cx from 'classnames';
 import { FormattedMessage as Msg } from 'react-intl';
+import { retrieveCall } from '../../../actions/call';
 
 import Avatar from '../../misc/Avatar';
 
-
+@connect()
 export default class CallListItem extends React.Component {
     static propTypes = {
         onItemClick: React.PropTypes.func.isRequired,
         data: React.PropTypes.object,
     };
+
+    componentDidUpdate(prevProps) {
+        const call = this.props.data;
+
+        if (this.props.inView && !prevProps.inView) {
+            if (!call || call.message_to_organizer !== undefined) {
+                return;
+            }
+            else if (call.organizer_action_needed) {
+                this.props.dispatch(retrieveCall(call.id))
+            }
+        }
+    }
 
     render() {
         let call = this.props.data;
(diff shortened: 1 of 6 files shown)


jensborje:
Would it be possible to include a screenshot? I can't test it myself on my local machine atm.

MikaelJohnAndersson:
@jensborje Sure! I updated the PR with added screenshots.

src/components/lists/items/CallListItem.jsx
```diff
@@ -1,16 +1,31 @@
...
 
+    componentDidUpdate(prevProps) {
+        const call = this.props.data;
+
+        if (this.props.inView && !prevProps.inView) {
+            if (!call || call.message_to_organizer !== undefined) {
+                return;
+            }
+            else if (call.organizer_action_needed) {
+                this.props.dispatch(retrieveCall(call.id, true))
```
martinajsongren: You're missing a semicolon on the end of this line. :)

MikaelJohnAndersson: Fixed by 241649f

src/components/lists/items/CallListItem.jsx
```diff
@@ -41,7 +57,34 @@ export default class CallListItem extends React.Component {
...
+            let messageContent = null;
+
+            if (call.message_to_organizer === null) {
+                messageContent = (
+                    <Msg tagName="span"
+                        id="lists.callList.item.organizerMsg.noMessage"/>
+                );
+            }
+            else if (call.message_to_organizer && call.message_to_organizer.length > 180) {
+                messageContent = call.message_to_organizer.substr(0, 180) + '...';
```
richardolsson: How about something like this?

```
let len = 0;
messageContent = call.message_to_organizer.split(' ').filter(s => (len += s.length + 1) && len < 40).join(' ') + '...';
```

src/components/lists/items/CallListItem.scss
```diff
@@ -185,3 +213,10 @@
         }
     }
 }
+
+@keyframes CallListItem-organizerMsg-loadingAnimation {
+    0%   { content: '.'; }
+    33%  { content: '..'; }
+    66%  { content: '...'; }
+    99%  { content: '.'; }
```
richardolsson: Is this line needed?

PR #1038, opened 2020-01-23
CallPane infinite dispatch loop bugfix

### LINKED ISSUE #1036: CallPane infinitely retrieving call assignment
`CallPane` infinitely retrieves call assignment while open, resulting in an infinite render loop.

MikaelJohnAndersson:
Fixes #1036

### DIFF
--- a/src/components/panes/CallPane.jsx
+++ b/src/components/panes/CallPane.jsx
@@ -70,8 +70,11 @@ export default class CallPane extends PaneBase {
         }
     }
 
-    componentWillReceiveProps(nextProps) {
+     componentWillReceiveProps(nextProps) {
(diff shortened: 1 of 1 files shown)


src/components/panes/CallPane.jsx
```diff
@@ -70,9 +70,12 @@ export default class CallPane extends PaneBase {
         }
     }
 
-    componentWillReceiveProps(nextProps) {
+     componentWillReceiveProps(nextProps) {
```
olleolleolle: Is this whitespace change necessary, or perhaps a typo?

MikaelJohnAndersson: Oops! Fixed by bede541

PR #1041, opened 2020-02-02
Fix SurveySubmission filtering bug

### LINKED ISSUE #1040: Filtering SurveySubmissions on linking status not working
When filtering SurveySubmissions on linking status the list should only contain results matching current filter.

Example: Filtering on linked submissions
Expected behavior: Returns linked submissions
Actual behavior: Returns unlinked submissions

martinajsongren:
Fixes issue where filtering on linked/unlinked SurveySubmission always return unlinked submissions.
Closes #1040

### DIFF
--- a/src/actions/surveySubmission.js
+++ b/src/actions/surveySubmission.js
@@ -8,7 +8,7 @@ export function retrieveSurveySubmissions(surveyId = null, linked = null, page =
         let promise;
 
         if (linked !== null) {
-            linked = linked? ((filters.linked == 'linked')? 1 : 0) : null;
+            linked = linked? ((linked == 'linked')? 1 : 0) : null;
             filters.push(['linked', '==', linked? 1 : 0]);
         }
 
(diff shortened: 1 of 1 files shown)


src/actions/surveySubmission.js
```diff
@@ -8,7 +8,7 @@ export function retrieveSurveySubmissions(surveyId = null, linked = null, page =
         let promise;
 
         if (linked !== null) {
-            linked = linked? ((filters.linked == 'linked')? 1 : 0) : null;
+            linked = linked? ((linked == 'linked')? 1 : 0) : null;
```
olleolleolle: As a formatting comment, when I write ternary conditional statements, I usually place the question mark whitespace-separated. 

The hope then is that the reader can separate the parts of the statement.

martinajsongren: I agree with this, but totally missed it when fixing this issue. Thanks for pointing it out! Will commit change asap :)

PR #1051, opened 2020-02-16
Feature/import client validation

### LINKED ISSUE #532: Warn when importing without ID
When trying to import without External ID and/or ZetkinID, prompt the user before proceeding. This is related to a general import confirmation (#523).

### LINKED ISSUE #570: Warn about too long values on import
All person data fields have maximum lengths, but in the interface it's possible to select a column type for a column that includes values that are too long. Upon import, this will cause an error (or perhaps in a future version of the platform, the values will be truncated) and the user will not know why. We should show a warning already in the importer interface, before importing.

### LINKED ISSUE #1049: Warn when trying to import invalid file
When trying to import a non-xlsx file, nothing happens. It would make sens to display a warning saying unsupported file format and which file formats are allowed.

samuelskanberg:
Doing some validation on e.g. what columns are chosen for doing import. We demand that first_name and last_name columns are chosen, no field chosen for more than one column and suggest to the user to use external id. The preview is simply saying that "X number of rows will be imported"

Resolves #532 
Resolves #570 
Resolves #1049


niklasva82:
There is a problem with the absolute requirement that first name and last name are present in an import. On an update, there is no need for first name or last name as long as the external id or zetkin id fields are present.

For instance, the Microsoft Dynamics export can sometimes contain only a "name"-field, not divided between first name and last name. This is not a problem as long as the external id field is present and the person already exists in Zetkin, but this validation would stop the file from being imported. This is from an actual scenario where two files were exported from MS Dynamics with different fields missing in each.

I think a better solution would be to make first name and last name optional but with a strong warning, similar to how external id is treated, barring a solution to fetch all external/zetkin id:s and comparing the list client side but that seems like overkill.

richardolsson:
Good point @niklasva82. Do you want to make that change?

niklasva82:
Superseded by #1052

PR #1052, opened 2020-03-10
Import async

### LINKED ISSUE #532: Warn when importing without ID

### LINKED ISSUE #570: Warn about too long values on import

### LINKED ISSUE #847: Handle import errors
There are several cases where file parsing can fail. Errors are currently not handled at all. Some examples:

* Wrong file format
* Password protected Excel files
* Too large files (>65535 rows)

### LINKED ISSUE #1043: Client-side validation of import
Before allowing an import to go ahead, validate the field selection.

* If no field has been selected as an external ID
   - Warn the user about this, but allow them to go ahead if they really want to.
* First name and last name
   - Check that the user has selected first name and last name
   - Check that all rows in the selected first name and last name columns are non-null
   - If one of these conditions are not fulfilled, tell the user about the problem and do not allow the user to go ahead with the import.

### LINKED ISSUE #1049: Warn when trying to import invalid file

niklasva82:
Accept new importer response and add a view for async import.

Implement client side validation
* Add validation of columns:
  * Check for duplicate columns
  * Check for at least one of external id or Zetkin id, or a combination of first name and last name
* Add validation of data:
  * Check for valid format of Zetkin id field (integer)
  * Check for valid format of email field
  * Check for valid format of phone number field (integers, +, -, space)
  * Check for valid format of gender field (f, m, o, or _)
  * Check for valid length of all fields
* Warn when:
  * No ID field was specified, or no first name and last name was specified.
  * Zetkin ID is specified, recommending External ID if the user is not specifically re-importing exported Zetkin ID:s
* Notify user
  * Number of rows to be imported

Resolves #532
Resolves #570
Resolves #1049
Resolves #847 
Resolves #1043


src/components/sections/people/ImportPane.jsx
```diff
@@ -59,24 +59,34 @@ export default class ImportPane extends RootPaneBase {
                     dispatch={ this.props.dispatch }/>
             );
         }
-        else if (stats) {
+        else if (stats && stats['status'] == "completed") {
```
richardolsson: Why `stats['status']` instead of `stats.status`?

src/store/importer.js
```diff
@@ -16,7 +16,7 @@ export default function importer(state = null, action) {
     else if (action.type == types.EXECUTE_IMPORT + '_FULFILLED') {
         return Object.assign({}, state, {
             importIsPending: false,
-            importStats: action.payload.data.data.report,
+            importStats: action.payload.data.data,
```
richardolsson: Since this no longer contains just stats (the "statistics" of the import), perhaps this should be renamed, along with the `stats` local variable in `ImportPane`. I'm thinking something along the lines of `importResponse` or `lastImportLog`.

richardolsson:
I get this warning trying to import a real-life file. It's great that the importer interface is catching this, but since it's not telling me which row or e-mail address is invalid, it's difficult for me to find and fix the issue in a file with 2000 lines.

image

How difficult would it be to add information about where the error(s) were found?

richardolsson:
Here are some more comments on functionality, having tested the new feature some more.

First of all, I think it looks great. Whatever worries I had about visuals are gone. All my concerns are about limitations concerning usage, having tested with some real-life data. Here's a checklist of what I feel needs to improve before we can merge this without causing disruptions for users:

- [ ] Include information about which rows are invalid (not just the first one, but all of them)
- [ ] Handle when cells are `undefined`
- [ ] Handle (stip) leading/trailing whitespace
- [ ] Catch when a row contains neither names nor ID
- [ ] Close confirm pane when cancelling import

More information about each of these below:

## Row numbers
When a row is invalid, the number should be displayed in the validation interface, so that the user knows what rows to fix.

If there are multiple errors, all of them should be displayed, so that the user doesn't have to retry over and over again after each fix (I had to retry maybe 15 times before my file, which was recently *exported from Zetkin*, was accepted).

## Handle `undefined` cells
Fields may be `undefined` which raises an exception in several cases (e.g. `row.values[i].length` when `row.values[i]` is `undefined`).
I found this to happen when I used a real life file where an e-mail address was undefined and external ID was undefined for multiple users. I'm not sure what causes the Excel parser to sometimes render a cell `undefined` and sometimes with an empty string value, but both need to be handled.

## Leading/trailing whitespace
Validation should be able to handle when an e-mail field contains leading/trailing whitespace. The best way to handle this would be to strip the whitespace before importing (so that the correct data, without whitespace, is imported). But at least it should not fail to recognize an e-mail address because it's followed by a space.

## Handle missing names and ID
The interface correctly warns if user does not configure at least an ID or first/last name columns. However, it does not seem to catch when the columns are correctly configured but not all lines fulfill the requirements.

I had a file which contained ID and names for most rows, but on one row contained only the ID, no names. This passed the validation, but failed on the back-end.

## Close confirm pane
When the import interface is cleared (using the "Abort import" button), the confirm pane is left open. This should be closed when the import is aborted. This could probably be achieved in the `view` reducer.

image

src/components/panes/ImporterColumnPane.jsx
```diff
@@ -13,7 +13,15 @@ import { resolveSettingsComponent } from '../misc/importer/settings';
 @connect(state => ({ importer: state.importer }))
 @injectIntl
 export default class ImporterColumnPane extends PaneBase {
-    getRenderData() {
+    shouldComponentUpdate(nextProps, nextState) {
+        if(nextProps.importer.tableSet == null) {
+            this.closePane();
```
richardolsson, line 18: This is fine, but in other cases it might be risky closing the pane when the data does not exist, because depending on the circumstances it might be temporary and the data might (re)appear shortly.

Instead, a more robust way would be to handle the relevant action (in this case `RESET_IMPORT`) in the `view` reducer and remove the pane there. No need to fix this now though, since this case should not be able to result in any bugs due to the special nature of the importer interface. Just an FYI for future reference.

src/components/sections/people/ImportPane.jsx
```diff
@@ -52,31 +53,53 @@ export default class ImportPane extends RootPaneBase {
...
         }
+        else if (parseError) {
+            return (
+                <div className="ImportPane-error">
+                    <Msg tagName="h1" id="panes.import.parseError.h"/>
+                    <Msg tagName="p" id="panes.import.parseError.p"/>
+                    <Button labelMsg="panes.import.error.resetButton"
+                        onClick={ this.onErrorResetButtonClick.bind(this) }
+                        />
+                </div>
```
richardolsson, line 64: Good, simple solution to handling bad files!

src/components/panes/ImporterColumnPane.jsx
niklasva82: Yeah, this was the simple solution that I found did not result in any errors due the pane closing and then trying to update stuff. I was looking into using the RESET_IMPORT reducer, however I did not find a simple way of finding the IDs of the open panes. I realized now that each pane has a type which I could filter on, so this has now been resolved and the view reducer is used instead.

src/components/misc/InfoList.jsx
```diff
@@ -13,13 +13,38 @@ export default class InfoList extends React.Component {
...
+                                }
+                                if(msgNumbers.length > max) {
+                                    msgValues[key] += this.props.intl.formatMessage( { id: 'misc.infoList.andMore' }, { number: (msgNumbers.length - max) })
+                                }
+                            }
+                        }
+                        value = this.props.intl.formatMessage({ id: item.msgId }, msgValues);
+                    }
+
+                    if(!value && item.datetime) {
```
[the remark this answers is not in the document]
niklasva82: This is used in the import_log branch. I will remove it for the time being.

src/components/misc/InfoList.jsx
```diff
@@ -13,13 +13,38 @@ export default class InfoList extends React.Component {
...
+                                    }
+                                    msgValues[key] += this.props.intl.formatNumber(msgNumbers[i]);
+                                }
+                                if(msgNumbers.length > max) {
+                                    msgValues[key] += this.props.intl.formatMessage( { id: 'misc.infoList.andMore' }, { number: (msgNumbers.length - max) })
+                                }
+                            }
+                        }
+                        value = this.props.intl.formatMessage({ id: item.msgId }, msgValues);
+                    }
```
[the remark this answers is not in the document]
niklasva82: This was to make the logic in ConfirmImportPane simpler. I figure this was more of a view handling issue and could be reused if another use-case comes up.

locale/misc/en.yaml
```diff
@@ -57,3 +57,5 @@ surveyOutline:
         saveBulkButton: Create {numLines, plural, =1 {one option} other {# options}}
 tagCloud:
     addButton: Add tag
+infoList:
```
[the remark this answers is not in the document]
niklasva82: I fixed this now.

PR #1058, opened 2020-04-03
CSV import

### LINKED ISSUE #1054: CSV file import
Implement a parser to allow CSV files to be imported using the person importer interface.

A user should be able to drop (or select) a CSV file in the importer the same way it's currently possible to drop (or select) Excel files. Once parsed everything should work exactly the same.

richardolsson:
This PR adds support for parsing and importing CSV files.

You can try with any CSV file, including most variations for delimiters, quotes et c. If the format is not recognized by the parser, a proper error is shown.

The CSV parser isn't quite as fancy as the Excel parser, but some of that is because the format itself is not as fancy. For example, the CSV parser will not look for a header (because there is no formatting) and hence will also not guess the contents of each column.

It will also not remove empty columns, which could theoretically be done but is not a prioritized feature and should probably be refactored into some general-purpose solution eventually.

Closes #1054

### DIFF
--- a/locale/panes/import/en.yaml
+++ b/locale/panes/import/en.yaml
@@ -33,7 +33,7 @@ error:
 parseError:
     h: Invalid file
     p: >-
-        The file you selected could not be loaded. Please convert your file to XLSX format and try again.
+        The file you selected could not be loaded. Please convert your file to CSV or XLSX format and try again.
 importButton: Import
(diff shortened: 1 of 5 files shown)


niklasva82:
Looks good! Properly parsed large real-world data converted to CSV.

PR #1060, opened 2020-04-04
Import log interface

### LINKED ISSUE #1053: Import log interface
Create an interface (e.g. a section) to render a list of previous imports, as introduced in pending changes to the Zetkin Platform.

Related to #1052.

niklasva82:
ImportLogList along with an ImportLogPane, listing preivous imports.

Resolves #1053


locale/panes/import/sv.yaml
```diff
@@ -106,3 +106,6 @@ tableSelect:
     p: |-
         Filen innehåller mer än ett ark. Välj vilket ark som innehåller
         den data du vill importera.
+viewSwitch:
+  import: Importera
+  logs: Historia
```
richardolsson, line 111: Maybe "Historik" is a better translation?

src/components/lists/items/ImportLogListItem.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+        return (
+            <div onClick={ this.onClick.bind(this) }className="ImportLogListItem">
+                <div className="ImportLogListItem-col">
+                    <FormattedDate value={ log.accepted } />
+                    <FormattedTime value={ log.accepted } />
+                </div>
+                <div className="ImportLogListItem-col">
+                    <Msg id={ "lists.importLogList.item.status." + log.status } />
+                    { log.report ? <FormattedNumber value={log.report.imported} /> : null }
```
richardolsson, line 31: I think this should be a label saying something like "15 rows", and not just the number 15.

src/components/lists/items/ImportLogListItem.scss
```diff
@@ -0,0 +1,61 @@
...
+    span {
+        font-size: 1.4em;
+        line-height: 1.5em;
+        display: block;
+        overflow: hidden;
+        text-overflow: ellipsis;
+    }
+
+    .ImportLogListItem-col {
+        @include small-screen {
```
richardolsson, line 19: I don't mind keeping this, but please spend no time on tweaking small-screen layouts of this since the import section is currently hidden on mobile.

src/components/misc/InfoList.jsx
```diff
@@ -39,6 +39,12 @@ export default class InfoList extends React.Component {
                                 }
                             }
                         }
+                        if(item.msgDatetimes) {
```
richardolsson, line 42: I don't mind this at this point, since it's already implemented, but the same thing should have been possible without changing `InfoList` by adding a `msgId` and the date in `msgValues`, and in the localization of the string using `{ myDate, date }` (where the `date` is a keyword telling the localization framework to format it as a date).

src/components/panes/ImportLogPane.jsx
```diff
@@ -0,0 +1,135 @@
...
+                ];
+                infoListData = infoListData.concat(completedData);
+            } else if(log.status == 'error') {
+                infoListData.push(
+                    {
+                        name: 'error',
+                        msgId: 'panes.importLog.error',
+                        value: log.error
+                    },
+                );
```
richardolsson, line 96: Is it really sensible to print out the error as stored in the database (line 94)? For one of my failed imports, this is what it looked like:

image

src/components/panes/ImportLogPane.jsx
```diff
@@ -0,0 +1,135 @@
...
+            /*
+            const avatar = log.imported_by ? (
+                <div className="ImportLogPane-imported_by">
+                    <Avatar key="avatar" ref="avatar" person={ log.imported_by } />
+                    <div className="ImportLogPane-imported_by-info"
+                        { log.imported_by.first_name + ' ' + log.imported_by.last_name }
+                    </div>
+                </div>)
+                : <Msg tagName="p" id="panes.importLog.noPerson" />;
+                */
```
richardolsson, line 108: Can this be removed?

src/components/panes/ImportLogPane.jsx
```diff
@@ -0,0 +1,135 @@
...
+                        <Avatar ref="avatar" person={ log.imported_by } />
+                        <div className="ImportLogPane-personInfo">
+                            <span className="ImportLogPane-name">
+                                { name }
+                            </span>
+                            <span className="ImportLogPane-email">
+                                { email }
+                            </span>
+                        </div>
+                    </div>;
```
richardolsson, line 124: It would be nice if this would link to the person pane. Maybe use an existing component (e.g. the ones used in `ActionPane` or `CallAssignmentPane`) for the person which should already support clicking including hover effects et c.

src/components/panes/ImportLogPane.scss
```diff
@@ -0,0 +1,88 @@
+.ImportLogPane {
+    .InfoListItem-status {
+        &:before {
+            @include icon($fa-var-question-circle);
```
richardolsson, line 4: How about `fa-var-check` when successful, `fa-var-exclamation` if it failed, `fa-var-hourglass-half` if pending?

src/components/panes/ImportLogPane.scss
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    .InfoListItem-accepted {
+        &:before {
+            @include icon($fa-var-question-circle);
```
richardolsson, line 10: How about `fa-var-upload`?

src/components/panes/ImportLogPane.scss
```diff
@@ -0,0 +1,88 @@
...
+
+    .InfoListItem-error {
+        &:before {
+            @include icon($fa-var-exclamation-triangle);
+        }
+    }
+
+    .InfoListItem-imported {
+        &:before {
+            @include icon($fa-var-cloud-upload);
```
richardolsson, line 28: This is actually (with some acceptable exceptions) the number of rows in the file, so I think the `fa-var-list` icon should work well here.

src/components/panes/ImportLogPane.scss
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    .InfoListItem-created {
+        &:before {
+            @include icon($fa-var-hand-paper-o);
```
richardolsson, line 34: How about `fa-var-plus`?

src/components/panes/ImportLogPane.scss
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    .InfoListItem-updated {
+        &:before {
+            @include icon($fa-var-recycle);
```
richardolsson, line 40: I think `fa-var-pencil` would be nice for this.

src/components/lists/items/ImportLogListItem.scss
```diff
@@ -0,0 +1,61 @@
...
+    position: relative;
+
+    &::after {
+        content: "";
+        display: block;
+        clear: both;
+    }
+
+    span {
+        font-size: 1.4em;
```
richardolsson, line 11: I think it would be nicer to use the default font-size. No other list has big labels like this.

src/components/misc/InfoList.jsx
niklasva82: Interesting. I tried this but it didn't work. It did work in another place with the <FormattedMessage> React-component for number formatting though. Could it be that it doesn't work with inject-intl?

richardolsson: It definitely should work, but lets not dive to deeply into it now. This is fine!

PR #1064, opened 2020-04-29
Gender import field

### LINKED ISSUE #350: Gender import field
Create special field for gender, mapping common (and configurable) values to the three genders.

niklasva82:
Implementation of mapping of the gender field on import.

Resolves #350

### DIFF
--- a/locale/panes/import/en.yaml
+++ b/locale/panes/import/en.yaml
@@ -83,6 +83,13 @@ settings:
     personTag:
         emptyLabel: Tags for empty rows
         valueLabel: 'Tags for rows with value: {value}'
+    personGender:
+        emptyLabel: Gender for empty rows
+        female: Female
+        male: Male
+        other: Other
+        unknown: Unknown
+        valueLabel: 'Gender for rows with value: {value}'
 summary:
     personTag:
         summary: |-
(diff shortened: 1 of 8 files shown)


niklasva82:
> This is looking great, but I think there is a misunderstanding about the intended values for the `gender` field.
> 
> They values that should be allowed are:
> 
>     * `"m"` for male
> 
>     * `"f"` for female
> 
>     * `"o"` for other
> 
>     * `null` when unknown
> 
> 
> The `"_"` value that is used in some places in the code exists to represent null in HTML elements (e.g. `<select>`) but the actual value sent to the API should be `null`. This PR sends it to the server, which is not intended behavior.
> 
> Note how it's handled in `PersonForm`:
> 
> https://github.com/zetkin/organize.zetk.in/blob/ec853937500a139f35c53738156d5a80ffe8ccba/src/components/forms/PersonForm.jsx#L54-L62
> 
> Obviously there is not enough data validation in the API since it is allowing this invalid data (`"_"`) to be stored, but until we implement proper validation there, we should at least not send invalid data "by design" from our interface.
> 
> Please convert `"_"` to `null` upon sending before we merge.

OK, fixed.

Issue #1067, opened 2020-06-05
Norwegian translations


richardolsson:
Add Norwegian localized strings.

olleolleolle:
Do you have the language strings, raw, anywhere?

richardolsson:
> Do you have the language strings, raw, anywhere?

Most of the missing strings are in our l10n editor software (poeditor) but the scripts to extract them have begun to rot a little bit. I think this requires a proper review of the i18n tooling before it can be completed.

richardolsson:
Was going to do this for R27 but realized that the translation hasn't been made yet. Will wait for the new [translator's interface](https://github.com/zetkin/translators-interface) before starting on this.

PR #1068, opened 2020-06-05
Norwegian locale

richardolsson:
Add Norwegian locale. There are no strings for Organize yet, but the locale needs to be recognized, and the English fallback used, which is the case with this PR.

Relevant to #1067


richardolsson:
Incorrect PR, accidentally branched off `production`.

PR #1075, opened 2020-06-23
Person (organizer) notes

richardolsson:
This PR adds interface to read and create organizer notes on person objects.

image


richardolsson:
Thanks! I agree about delete buttons, but I think that can wait until we do an (eventually necessary) re-design of this feature.

I've created a new issue for this, #1076

PR #1078, opened 2020-07-05
Hide custom fields in PersonPane

### LINKED ISSUE #1072: Hide fields from PersonPane when unused
When the custom fields feature is unused by an organization, i.e. when there are no fields, consider hiding that section from `PersonPane` altogether.

niklasva82:
Resolves #1072


richardolsson (approved):
LGTM

PR #1080, opened 2020-08-22
Fix action import bug caused by CSV import feature

### LINKED ISSUE #1079: Action import fails in recent versions
The CSV import feature added recently has introduced a bug in the campaign/action import, which now receives a promise but is expecting it to be a resolved object.

Steps to reproduce
1. Open the action import pane from the calendar root pane
2. Drag any Excel file to the drop zone

Expected results: File should be parsed and either show a list or some error message
Actual results: Nothing happens in UI, error is thrown in console:

```
main.js:47 Uncaught (in promise) TypeError: Cannot read property 'items' of undefined
    at actionImport (main.js:47)
    at main.js:66
    at p (<anonymous>:1:36402)
    at v (<anonymous>:1:36684)
    at <anonymous>:1:40069
    at Object.dispatch (main.js:36)
    at e (<anonymous>:1:40553)
    at main.js:47
    at main.js:47
    at main.js:66
```

This seems to be because src/utils/importer.js now returns a promise and not a resolved "table set" object.

Proposed fix: Make sure that the `parseActionImportFile()` action handles the returned promise from `parseWorkbook()` instead of incorrectly assuming it's a plain object.

niklasva82:
Resolves #1079

### DIFF
--- a/src/actions/actionImport.js
+++ b/src/actions/actionImport.js
@@ -10,8 +10,9 @@ export function parseActionImportFile(file) {
 
         reader.onload = e => {
             // TODO: Check type and support CSV as well
-            let tableSet = parseWorkbook(e.target.result);
-            resolve({ tableSet });
+            parseWorkbook(e.target.result)
+                .then(tableSet => resolve({ tableSet }))
+                .catch(error => reject(error));
         };
 
         reader.readAsBinaryString(file);
--- a/src/actions/actionImport.js
+++ b/src/actions/actionImport.js
@@ -11,7 +11,7 @@ export function parseActionImportFile(file) {
         reader.onload = e => {
             // TODO: Check type and support CSV as well
             parseWorkbook(e.target.result)
-                .then(tableSet => resolve(tableSet))
+                .then(data => resolve(data))
                 .catch(error => reject(error));
         };
 
--- a/src/actions/actionImport.js
+++ b/src/actions/actionImport.js
@@ -11,7 +11,7 @@ export function parseActionImportFile(file) {
         reader.onload = e => {
             // TODO: Check type and support CSV as well
             parseWorkbook(e.target.result)
-                .then(tableSet => resolve({ tableSet }))
+                .then(tableSet => resolve(tableSet))
                 .catch(error => reject(error));
         };
 
--- a/src/store/actionImport.js
+++ b/src/store/actionImport.js
@@ -5,7 +5,7 @@ import makeRandomString from '../utils/makeRandomString';
 
 export default function actionImport(state = null, action) {
     if (action.type == types.PARSE_ACTION_IMPORT_FILE + '_FULFILLED') {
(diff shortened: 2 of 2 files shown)


src/actions/actionImport.js
```diff
@@ -10,8 +10,9 @@ export function parseActionImportFile(file) {
 
         reader.onload = e => {
             // TODO: Check type and support CSV as well
-            let tableSet = parseWorkbook(e.target.result);
-            resolve({ tableSet });
+            parseWorkbook(e.target.result)
+                .then(tableSet => resolve({ tableSet }))
```
olleolleolle, line 14: If you passed the bare `tableSet`, not wrapped in a `{ tableSet: tableSet }` literal, could you do without the  actionImport.js change, then?

olleolleolle (commented):
I looked at the code, briefly, 👏 and yay it looks like it fixes that Issue on sight, but I wonder if it is possible to make the change even smaller.

src/actions/actionImport.js
richardolsson: @olleolleolle The `parseWorkbook()` function is also used [elsewhere](https://github.com/zetkin/organize.zetk.in/blob/master/src/actions/importer.js), where it's expected to return an object like this. If I understand your suggestion correctly, that would break (or require more changes). Am I right?

olleolleolle: My suggestion came from: the _other_ change in this PR, is that a follow-up, un-related fix to this repair?

niklasva82: action/actionImport is not used elsewhere to my knowledge, I suppose I could just resolve(tableset), then there's no need to change store/actionImport and it would actually be better from a compatibility standpoint if it was used elsewhere for some odd reason.

olleolleolle: Thanks for the follow-up questions, Richard. And, I hope this comes across as my trying to be understood, rather than anything else. All this is Good Work.

richardolsson: Ah, I get it now. I was confused by the naming i guess. The first use of `tableSet` here, i.e. the argument in the function is not actually the table-set, but an object containing the field `tableSet` I believe? So maybe a better name would be `data` or `workbook` or something, along with the change suggested by Olle?

I'll let you decide @niklasva82 since I might be mistaken. Do you want to update the PR before I review?

niklasva82: @richardolsson True regarding the naming. I fixed that as well now.

olleolleolle (approved):
❤️ 💯

richardolsson (commented):
Thanks for simplifying, and thanks to @olleolleolle for keen eyes and suggestions!

The code looks good. I will try running it as well before merging but this should be fine!

richardolsson (approved):
Tested and found to be working!

PR #1082, opened 2020-08-29
Hide campaign visibility select when campaign is not published

### LINKED ISSUE #846: Hide visibility when not published
When a campaign is a draft (not published), the `visibility` attribute makes no sense. Hide it in the `CampaignForm` when `published == false`.

richardolsson:
Fixes #846


olleolleolle (approved):
Looks good to me, on reading

PR #1083, opened 2020-08-29
Added link to edited campain

### LINKED ISSUE #1044: Add link to public campaign page on campaign pane
The public campaign page can be difficult to find for users who do not know where to look. Add a link to https://zetk.in/o/{organization id}/campaigns/{campaign id} to campaign pane

fannysoderlund:
Fixes #1044

### DIFF
--- a/src/components/panes/EditCampaignPane.jsx
+++ b/src/components/panes/EditCampaignPane.jsx
@@ -1,13 +1,13 @@
 import React from 'react';
-import { injectIntl } from 'react-intl';
-import { connect } from 'react-redux';
+import {injectIntl} from 'react-intl';
+import {connect} from 'react-redux';
 
 import PaneBase from './PaneBase';
 import CampaignForm from '../forms/CampaignForm';
 import Button from '../misc/Button';
 import LoadingIndicator from '../misc/LoadingIndicator';
-import { getListItemById } from '../../utils/store';
-import { retrieveCampaign, updateCampaign, deleteCampaign }
+import {getListItemById} from '../../utils/store';
+import {retrieveCampaign, updateCampaign, deleteCampaign}
     from '../../actions/campaign';
 
 const mapStateToProps = (state, props) => ({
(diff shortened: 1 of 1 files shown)


src/components/panes/EditCampaignPane.jsx
```diff
@@ -41,6 +42,17 @@ export default class EditCampaignPane extends PaneBase {
         }
     }
 
+    getPaneSubTitle(data) {
+        if (this.props.campaignItem && !this.props.campaignItem.isPending) {
+            const link= 'https://zetk.in/o/'+this.props.activeOrg.id+'/campains/'+this.props.campaignItem.data.id;
```
richardolsson, line 47: This hardcodes the URL to the domain `zetk.in` using HTTPS, which is correct for the live environment but not correct for the development environment. Instead of https, use `//` (same scheme as current page) and instead of the domain, use `process.env.ZETKIN_DOMAIN`.

See _Dashboard.jsx_ for a similar solution:
https://github.com/zetkin/organize.zetk.in/blob/feacfdfba5d2f08c39b6c03e414e90d10c2d39f6/src/components/dashboard/Dashboard.jsx#L65

src/components/panes/EditCampaignPane.jsx
```diff
@@ -41,6 +42,17 @@ export default class EditCampaignPane extends PaneBase {
...
+
+            return <a href={link} target="_blank"> {link}</a>;
```
richardolsson, line 49: There is a redundant whitespace on this line before the second `{link}` which should be removed.

src/components/panes/EditCampaignPane.jsx
```diff
@@ -41,6 +42,17 @@ export default class EditCampaignPane extends PaneBase {
...
+      }
```
richardolsson, line 50: This is incorrectly indented. Should be 8 spaces, but is currently just 6.

richardolsson (changes requested):
Great work on your first pull request!



This works (almost) as expected, but I do have some minor changes that I would like to ask you to make. See comments below!

richardolsson:
Superseded by #1088 which contains some additional fixes.

PR #1084, opened 2020-08-29
Use SSH to clone, not HTTPS

oelrich:
Very minor edit. Also a .gitattributes that should take care of LF/CRLF-convertion problems.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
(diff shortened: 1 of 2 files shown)


richardolsson:
This is what we discussed @olleolleolle! I forgot that @oelrich had already suggested the change.

Olle, will you take a look and maybe update the developer tutorial accordingly (on how to set up SSH)?

olleolleolle:
@richardolsson Thanks for the pointer!

https://github.com/zetkin/developer.zetkin.org/pull/15 now mentions the "Connecting to GitHub with SSH" article

@oelrich Hey, good call with the gitattributes file!

olleolleolle (approved):
This is a useful configuration change, esp. for Windows developers, for whom changed line endings could be troublesome.

PR #1089, opened 2020-09-13
Feature/action import title support

### LINKED ISSUE #1081: Title in action import
The action importer feature does not support the recently added `title` field for actions. It should be added:

- [ ] Add title support to the parser
- [ ] Add title column to the instructions in the importer pane
- [ ] Add title to the data sent to the API

samuelskanberg:
Resolves #1081

### DIFF
--- a/locale/panes/importActions/en.yaml
+++ b/locale/panes/importActions/en.yaml
@@ -30,6 +30,7 @@ dropZone:
         location: Name of location
         participants: Minimum participant count (a single number, e.g. 2 or 11)
         startTime: Start time (24 hour format with or without minutes, e.g. 7:00, 13.00 or 16)
+        title: Title of this specific action
     h: Drop file here
     p: |-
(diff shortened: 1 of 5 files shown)


richardolsson (approved):
Nice work! Looks good and works as expected 👍

richardolsson:
Actually, on second thought, should we switch places for the title and description? What do you think, @samuelskanberg and @niklasva82?

niklasva82:
> Actually, on second thought, should we switch places for the title and description? What do you think, @samuelskanberg and @niklasva82?
> 
> Edit: For clarification, I mean switch columns in the spreadsheet, so that title comes before (to the left of) the description text in the spreadsheet format expected by the importer (and hence in the instructions).

The reason for this choice was backwards compatibility, so that if someone used this function previously and simply reuses their previous file with changes. This way, everything will be in its proper place in this case, whereas if we switch places of the two, description will become title.

PR #1091, opened 2020-09-13
Add darker green bg color to indicate that an action is splitable

### LINKED ISSUE #370: Indicate when participants are twice the required number
In legacy Zetkin, when an action has twice the number of required participants, the green indicator in the action list changes to a darker shade. This is to indicate that there are enough participants for the action to be split into two actions.

Lets add this in at some point.

samuelskanberg:
Resolves #370


richardolsson:
This is good, but @fannysoderlund worked on the same task unfortunately, and I like her solution slightly better. I would love to see you go ahead with your suggestion for a label though!

Closing in favor of #1092.

PR #1092, opened 2020-09-13
Indicate 2x participant count with darker color

### LINKED ISSUE #370: Indicate when participants are twice the required number

fannysoderlund:
Fixes #370


richardolsson (changes requested):
I like this! I would like it slightly brighter, like 30% darkened. Other than that I'm ready to merge!

richardolsson (approved):
LGTM! This will make it easier for organizers to tell instantly when an action is overbooked enough that it would make sense to split it into two actions. Good work!

PR #1094, opened 2020-09-26
Fix error when bulk deleting people

### LINKED ISSUE #937: Exception in bulk delete
Trying to delete several people using the bulk delete feature throws an unhandled promise rejection. Steps to reproduce:

1. Navigate to the people list
2. Select 3 person objects
3. From the bulk op list, pick "Delete 3 people" and click "Perform operation"

Expected results: The button should show a pending state, and the pane should then close with the people removed
Actual results: The button shows a pending state indefinitely. The console prints out an unhandled error warning.

### LINKED ISSUE #1007: Bulk delete bug
This happens when trying to delete a bulk selection of people, probably due to type inconsistencies for the organization ID (re-sending data with int ID works).

```
zetkinorganize_1     | 2019-05-06T09:04:18.240836412Z Error executing bulk op person.delete
zetkinorganize_1     | 2019-05-06T09:04:18.241718630Z Trace: TypeError: Cannot read property 'profile' of undefined
zetkinorganize_1     | 2019-05-06T09:04:18.241755420Z     at req.z.resource.get.then.result (/var/app/build/app/server/api/bulk.js:139:89)
zetkinorganize_1     | 2019-05-06T09:04:18.241776907Z     at process._tickDomainCallback (internal/process/next_tick.js:135:7)
zetkinorganize_1     | 2019-05-06T09:04:18.241793116Z     at exec.then.catch.err (/var/app/build/app/server/api/bulk.js:35:21)
zetkinorganize_1     | 2019-05-06T09:04:18.241809584Z     at process._tickDomainCallback (internal/process/next_tick.js:135:7)
```

richardolsson:
Fixes #937 
Fixes #1007

This seems to have been introduced by #984, which prevents users from deleting themselves by checking their memberships and filtering their profile out of the people that are about to be deleted.

The issue is that organization ID is sometimes(?) sent as a string and sometimes as an integer, but was compared using strict equality (`===`). This meant that sometimes no matching membership would be found, causing an exception.

Just switching to `==` means that a membership will always be found. If for some reason it's not, this should fail anyway.

### DIFF
--- a/src/server/api/bulk.js
+++ b/src/server/api/bulk.js
@@ -161,7 +161,7 @@ let operations = {
             .get()
             .then(result => {
                 const userPersonId = result.data.data
-                    .find(m => m.organization.id === orgId)
+                    .find(m => m.organization.id == orgId)
                     .profile.id;
 
                 const promises = req.body.objects


richardolsson:
This fixes the issue I was having which you couldn't reproduce, @oelrich. I'm not sure why it was inconsistent, but at least this solves the issue.

PR #1095, opened 2020-09-26
Implement 2-step interaction in DeleteButton

### LINKED ISSUE #302: 2-step delete buttons
Right now, delete buttons in Organize delete instantly. Instead, there should be some type of confirmation. Perhaps similar to how deleting tags work on GitHub?

__Step 1 (default state)__:
image

__Step 2 (having clicked delete)__:
image

johnschack:
Implement 2-step interaction when deleting, for example an action, to avoid deleting by error. See screenshot below.

Before clicking
image

After clicking (confirm and cancel clickable)
image

Fixes #302

### DIFF
--- a/locale/misc/en.yaml
+++ b/locale/misc/en.yaml
@@ -21,6 +21,9 @@ campaignSelect:
     nullLabel: All campaigns
 deleteButton:
     label: Delete
+    confirm: Confirm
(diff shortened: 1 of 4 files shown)


richardolsson:
I'd be curious to hear your thoughts on this, @jensborje!

PR #1098, opened 2020-09-29
Fix incorrect field name for zip code in PersonDataFilter

richardolsson:
This PR fixes an undocumented issue with the name of the zip code field in `PersonDataFilter`. It was named `zip` but is expected by the Zetkin platform API to be named `zip_code`.

Steps to reproduce (before applying this fix):

1. Create a new Smart Search Query, name it whatever
2. Add a new Person Data filter
3. Set the field to search for to Zip
4. Type in any Zip
5. Save the query

Expected result: API should accept input
Actual result: API rejects input

### DIFF
--- a/src/components/filters/PersonDataFilter.jsx
+++ b/src/components/filters/PersonDataFilter.jsx
@@ -28,7 +28,7 @@ export default class PersonDataFilter extends FilterBase {
             'phone': 'filters.personData.fields.phone',
             'co_address': 'filters.personData.fields.co_address',
             'street_address': 'filters.personData.fields.street_address',
-            'zip': 'filters.personData.fields.zip',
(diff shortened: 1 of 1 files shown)


niklasva82:
Getting the same error as in the test.
{"error": {"title": "Invalid filter format", "description": ""}}

niklasva82 (commented):
Getting the same error as in the test on core when following the instructions. core is on bugfix/person-data-filter-validation

{"error": {"title": "Invalid filter format", "description": ""}}

richardolsson:
@niklasva82: Did you restart core after switching branches?

richardolsson:
I have tested it again and it works for me. Merging for today's release.

PR #1121, opened 2020-10-21
Feature/custom fields import

### LINKED ISSUE #1096: Custom field import
Add column types to importer interface to import data to custom fields.

niklasva82:
Implements UI for custom field import.

Resolves #1096

### DIFF
--- a/src/actions/importer.js
+++ b/src/actions/importer.js
@@ -104,6 +104,11 @@ export function updateImportColumn(tableId, columnId, props) {
                 }
 
                 props.config = { mappings };
+            } else if (props.type == 'person_field') {
+                props.config = {
+                    field_id: props.config.field_id,
+                    field_type: props.config.field_type,
+                }
             }
         }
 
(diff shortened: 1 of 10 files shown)


niklasva82:
There is a problem with the select box that I haven't been able to track down. Everything seems to work properly otherwise, it's a cosmetic issue but pretty important from a users point of view.
To reproduce:
1. Select a file to import
2. In one of the columns, select a custom field

Expected result: The select box stays at that field.
Actual result: The select box jumps back to "Skip column", even though the state changes in the background and the import including validation still works.

src/actions/importer.js
```diff
@@ -139,23 +144,39 @@ export function executeImport(tableId) {
...
+                        if(m.value === value) {
+                            return true;
+                        }
+                    });
+                    r[genderIdx] = new_value && new_value.gender ? new_value.gender : null;
+                }
+                // Parse JSON column from text to structured JSON
+                for(const jsonIdx of jsonFieldIdxs) {
+                    let value = r[jsonIdx];
+                    r[jsonIdx] = JSON.parse(value);
```
richardolsson, line 171: This is not dealing correctly with empty cells. I tried to find an easy solution (e.g. `!!value? JSON.parse(value) : null`) but I think that only introduces other issues, like how do we deal with accidental overwriting when cells are empty? There also seems to be some issues with how arrays are handled when sent off to the back-end.

My suggestion is that we don't support JSON fields at all in the importer. I.e. filter them out in `ImporterColumnHead` and remove the validation/parsing code.

src/components/misc/importer/ImporterColumnHead.jsx
```diff
@@ -36,6 +38,11 @@ export default class ImporterColumnHead extends React.Component {
             'person_tag': col('colOptions.complex.tag'),
         };
 
+        const FIELD_OPTIONS = this.props.fieldTypes.items.reduce((obj, field) => {
+            obj['person_field.' + field.data.id + '.' + field.data.type] = field.data.title;
```
richardolsson, line 42: This seems unnecessarily complex. Maybe this is what's causing that bug you were experiencing? I'm guessing this is there to be able to send `field_type` along with the `config` but as I've mentioned elsewhere I don't think we should be including `field_type` there. Which should allow for simplifying this.

richardolsson (changes requested):
I know this is still a draft, but since we seem to miss each other during the days I thought I'd still do an initial review.

As far as I can tell, it seems to work as expected. But I think it's unnecessarily complex in some ways. One such way is JSON support, which I think we could do without (it's enough to support JSON through the API).

Another way is the `field_type` configuration attribute, which I think we should get rid off and instead let the back-end figure out the type based on `field_id`.

Let me know what you think!

richardolsson:
On a separate note, I think we should put the _Custom Fields_ section right after the _Personal data fields_ section in the `<select>`, before the _Other data_ section. Custom fields and Person fields are closely related. Do you agree?

image

niklasva82:
> On a separate note, I think we should put the _Custom Fields_ section right after the _Personal data fields_ section in the `<select>`, before the _Other data_ section. Custom fields and Person fields are closely related. Do you agree?
> 
> image

Makes sense, fixed.

richardolsson:
Is this ready for review now? @niklasva82

niklasva82:
Fixed the UI bug and it's now ready for review.

richardolsson (approved):
LGTM! 🎉

PR #1124, opened 2020-10-21
Retrieve field types in PersonPane

richardolsson:
Fixes undocumented bug which prevented custom fields from being displayed.

This bug was introduced in #1078, which only retrieves the custom field types in the `PeopleListPane`, but that pane doesn't always precede the `PersonPane`. The bug could be reproduced as such:

1. Refresh Organize on the home page with dummy org 1 (to reset state)
2. Navigate to Campaign section
3. Search for a person
4. Pick any of the search results – `PersonPane` opens over campaign section

Expected results: `PersonPane` should contain info about custom fields
Actual results: `PersonPane` does not contain the custom fields section because it does not find any field types


niklasva82 (approved):
Works as advertized!

PR #1126, opened 2020-10-22
In-view tag toggling

### LINKED ISSUE #1112: Toggle tag from view
Add support for toggling a tag using the checkbox in a view row.

richardolsson:
This PR builds on the recent Views feature to add the ability to set/unset tags directly in tag columns.

image

To achieve this, a framework has been created in the view reducer which allows for handling actions that change information related to people by flagging relevant rows in affected views as `dirty`. When the view components encounter a dirty row, it will trigger a re-fetch of the row before rendering.

This means that the framework also updates rows even if the change was triggered elsewhere, e.g. by the tag cloud in `PersonPane`. It's easy to extend this going forward to allow for editing other information in view rows, and trigger updates, e.g. personal information.

Closes #1112


src/store/personViews.js
```diff
@@ -226,11 +289,118 @@ export default function personViews(state = null, action) {
...
-            rowsByView: {},
-        };
+        // Not a view related action, but it could still be an action that
+        // affects data in a view, and should flag a row as dirty. Check
+        // for action types which should have this affect, and create a list
+        // of dirtyPersonIds in dirty views.
+        const dirtyPersonIds = [];
+        const affectedViewIds = [];
+
+        if (action.type == types.ADD_TAGS_TO_PERSON + '_FULFILLED') {
```
niklasva82, line 299: I suppose we could add more types here later, e.g. person data or custom fields?

src/components/misc/personViews/PersonViewTableRow.jsx
```diff
@@ -7,8 +8,13 @@ import BooleanViewCell from './cells/BooleanViewCell';
...
 import SurveyResponseCell from './cells/SurveyResponseCell';
 
+import {
+    addTagsToPerson,
+    removeTagFromPerson,
+} from '../../../actions/personTag';
 
-export default function PersonViewTableRow(props) {
+
+export default connect()(function PersonViewTableRow(props) {
```
niklasva82, line 17: Is there a reason for not using the decorator here?

niklasva82 (approved):
Looks great! I had a couple of questions but nothing that impacts my approval.

src/store/personViews.js
richardolsson: Yes exactly, as hinted in the PR description. 👍

src/components/misc/personViews/PersonViewTableRow.jsx
richardolsson: I don't know why but I just assumed that a function declaration like this can't be "decorated". Don't know why I assumed that.

richardolsson: I tried changing it now and it turns out my assumption was correct, at least with the version of babel that we are using. So I won't make any changes to this.

PR #1127, opened 2020-10-22
Fix duplicate person_field bug

niklasva82:
There is a bug where two columns of the same person_field data-type, but with different IDs, triggered a duplicate column error in the client-side validation. This pull request fixes that. The solution, with a regex-match in a switch-statemnt, is not very pretty, but rewriting the whole thing with a dozen if-statements is not very pretty either.


richardolsson (approved):
Works as expected! 💯 

As you've mentioned this is not a very beautiful fix, but it's because the framework is not very beautiful. I have created a separate issue #1128 to refactor the framework later.

PR #1131, opened 2020-10-23
Lazy-loading Avatar

### LINKED ISSUE #1129: Lazy load avatars
Viewing a large list of people causes all the avatars to load instantly, which can be a huge performance issue when the list contains hundreds or thousands of people. Pagination is great in some cases, but not always (and not as a means to prevent off-screen avatars from loading).

The [new ``](https://web.dev/browser-level-image-lazy-loading/) HTML attribute would let the browsers solve this. At the time of writing, [roughly 75% of worldwide users](https://caniuse.com/loading-lazy-attr) use browsers that would respect this property.

However, the version of `react-dom` that we are using in Organize does not understand this property and will strip it from the rendered HTML.

I can think of two solutions:

* Use a modified version of react-dom hosted on our GitHub and installed from there
* Set the attribute at runtime after mounting. Might cost more than it saves, and might not solve the problem in case the browser already started loading?

More information in [this StackOverflow question](https://stackoverflow.com/questions/31273093/how-to-add-custom-html-attributes-in-jsx).

I have also tried accessing the `HTMLDOMPropertyConfig` internal React object from the outside, e.g. from `App.jsx`, to add the property at runtime, but I can't seem to get access to it in any neat way.

richardolsson:
This PR sets the "new" `loading="lazy"` attribute on the `` element in `Avatar`, as discussed in #1129. Because it can't be done through the declarative React JSX (because our version of `react-dom` is too old) I've done it using the regular DOM API post mount. It seems to work just fine without performance issues.

The PR also sets avatars to always be square, regardless of width, using the `padding-top` CSS trick. This was always a bit of an issue causing avatars to flicker as they loaded, but it became even more obvious when lazy-loaded.

Fixes #1129

### DIFF
--- a/src/components/misc/Avatar.jsx
+++ b/src/components/misc/Avatar.jsx
@@ -50,6 +50,14 @@ export default class Avatar extends React.Component {
         if (this.refs.image.complete) {
             this.onLoad();
         }
+        else {
+            // Ideally we'd like to set loading="lazy" in the JSX inside
+            // render(), but it does not work with our version of react-dom.
+            // Instead we have to do it "at runtime" in the browser using
+            // the regular DOM API.
+            // TODO: After upgrading react-dom to 16+, move to JSX
+            this.refs.image.setAttribute('loading', 'lazy');
+        }
     }
 
     onLoad() {
--- a/src/components/misc/Avatar.scss
+++ b/src/components/misc/Avatar.scss
@@ -1,13 +1,19 @@
 .Avatar {
+    position: relative;
     display: inline-block;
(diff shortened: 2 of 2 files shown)


niklasva82 (approved):
It looks good and works for when moving between views, but as I'm sure you're aware, lazy attribute is not added on the server-side pre-rendering, which means that the lazy attribute is only set on dynamically added avatars. I suppose this issue is difficult to solve without upgradign react or modifying react-dom as you described in #1129 though, so this seems like a good enough solution for now.

PR #1133, opened 2020-10-25
Better date parsing in import

### LINKED ISSUE #1128: Refactor importer column type handling
The bug fix provided in #1127 is ugly because the underlying code is ugly. We should refactor the field management to more elegantly handle the fact that we now have a much larger set of field types than we used to.

niklasva82:
Better date parsing in import, including parsing birthdays from ID numbers. Add DateColumnValue class so that users can see what date will be interpreted from the imported data.

Also fixes #1128

### DIFF
--- a/src/components/misc/importer/values/DateColumnValue.jsx
+++ b/src/components/misc/importer/values/DateColumnValue.jsx
@@ -11,8 +11,8 @@ export default function DateColumnValue(props) {
             );
         }
         return (
-            <span><Msg id={ 'panes.import.settings.date.invalid' } /></span>
+            <Msg id="panes.import.settings.date.invalid" />
         );
     }
-    return <span></span>;
+    return null;
 }
--- a/src/components/misc/importer/values/DateColumnValue.jsx
+++ b/src/components/misc/importer/values/DateColumnValue.jsx
@@ -1,0 +1,18 @@
+import React from 'react';
+import { FormattedMessage as Msg } from 'react-intl';
+import getDate from '../../../../utils/getDate'
+
+export default function DateColumnValue(props) {
+    if(props.value) {
+        const date = getDate(props.value);
+        if(date) {
+            return (
+                <span>{ date }</span>
+            );
+        }
+        return (
+            <span><Msg id={ 'panes.import.settings.date.invalid' } /></span>
+        );
+    }
+    return <span></span>;
+}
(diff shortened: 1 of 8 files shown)


src/components/misc/importer/values/DateColumnValue.jsx
```diff
@@ -0,0 +1,18 @@
...
+export default function DateColumnValue(props) {
+    if(props.value) {
+        const date = getDate(props.value);
+        if(date) {
+            return (
+                <span>{ date }</span>
+            );
+        }
+        return (
+            <span><Msg id={ 'panes.import.settings.date.invalid' } /></span>
```
richardolsson, line 14: This will return a span within a span. Why not just return the `Msg` directly?

src/components/misc/importer/values/DateColumnValue.jsx
```diff
@@ -0,0 +1,18 @@
...
+        );
+    }
+    return <span></span>;
```
richardolsson, line 17: Why return an empty span, and not `null`?

src/components/panes/ConfirmImportPane.jsx
```diff
@@ -25,17 +26,17 @@ export default class ConfirmImportPane extends PaneBase {
...
-            return item.data.type;
+            return { type: item.data.type };
         } else if (item.data.type == "person_field") {
             const field = this.props.fieldTypes.items.find((f) => 
                 f.data.id == item.data.config.field_id);
-            return field.data.type + '.' + item.data.config.field_id;
+            return { type: field.data.type, name: 'field.' + item.data.config.field_id };
         } else {
-            return "unknown_type";
+            return { type: "unknown_type" };
```
richardolsson, line 39: Nice solution! Would you say that this fixes #1128?

richardolsson (commented):
This is looking good but I have a few questions regarding the code that may or may not result in changes. Let me know your thoughts!

src/components/panes/ConfirmImportPane.jsx
niklasva82: Yes it does

src/components/misc/importer/values/DateColumnValue.jsx
niklasva82: Fixed

src/components/misc/importer/values/DateColumnValue.jsx
niklasva82: Fixed

richardolsson (approved):
Looked at the code, verified the functionality. All is well!

PR #1136, opened 2020-10-27
Fix an undocumented layout issue in JoinSubmissionPane

richardolsson:
## Before
image

## After
image


niklasva82 (approved):
Looks good!

PR #1137, opened 2020-10-27
View tools (export, search, pagination)

### LINKED ISSUE #1108: View export
Consider adding a feature to export a view (as introduced by #1103) in it's current state as an Excel spreadsheet.

### LINKED ISSUE #1116: View pagination
Implement (client-side) pagination of views. The client store can easily handle thousands of rows, but the React UI cannot. Preliminary tests show acceptable UX with 500 rows visible at a time.

### LINKED ISSUE #1117: View filtering
Implement search/filter in Views UI. Browser based "search in page" was gonna cut it, but won't if we paginate the views (as requested in #1116).

richardolsson:
This PR introduces a new "toolbar" to Views, adding search, pagination and export capabilities.

* As-you-type search for plaintext cells in the view (does not support survey responses at this point)
* Pagination with a 500 row page limit, which is necessary for performance
* Export entire view (all pages) as CSV file

image

Closes #1108 
Closes #1116
Closes #1117

### DIFF
--- a/locale/misc/pageSelect/en.yaml
+++ b/locale/misc/pageSelect/en.yaml
@@ -1,0 +1,1 @@
+label: Page {page} of {count}
--- a/locale/misc/personViewTable/en.yaml
+++ b/locale/misc/personViewTable/en.yaml
@@ -1,4 +1,9 @@
 tools:
+    count:
+        default: Showing {visible} rows
+        filtered: Showing {visible} rows ({total} total)
+        filteredPaginated: Showing {visible} of {matches} rows ({total} total)
+        paginated: Showing {visible} of {total} rows
     downloadButton: Download .csv
     search:
         placeholder: Type to filter rows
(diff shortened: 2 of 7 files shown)


richardolsson:
Before code review I'm going to ask @jensborje what he thinks about the UX.

richardolsson:
I have spoken to Jens and made som minor changes to UX. Ready for code review now!

niklasva82 (approved):
Code looks good and everything works.

One thing which I'm not sure about, downloading while searching results in the entire dataset, not the filtered dataset. I think the user would expect the latter but I'm not sure.

richardolsson:
> One thing which I'm not sure about, downloading while searching results in the entire dataset, not the filtered dataset. I think the user would expect the latter but I'm not sure.

Me and Jens discussed this at some length and finally decided to keep it this way. Lets see what the users say, and we can potentially change it in the next release.

Issue #1140, opened 2020-10-28
Complete Revamp (proposal)


richardolsson:
This is a proposal to revamp Organize entirely, from scratch. In that way, it's similar to zetkin/www.zetk.in#277.

## TLDR
We (mostly myself) began working on Organize in early 2015, almost six years ago. Since then, the team has grown and together we have learned so much from actual use of the product, and relevant technologies have evolved.

I believe a complete revamp is the way to go both for technical reasons and to improve UX. Examined separately they are probably good candidates for incremental improvements, but together they warrant a complete revamp.

### Stack/libraries
While the choice to go with React/Flux was the right one, it was still very early when that choice was made in late 2014 (React 0.12). So much as changed in the React ecosystem and concerning best practices since then. We have tried to stay up to date but mostly failed, and to migrate the current code base (React 15) to React 17 would be a huge hassle.

On the server side, we've upgraded Node sporadically, but always remained a few versions behind the current LTS version. Some upgrades have resulted in bugs.

### Language
At the time when we started development, JS/ES was evolving extremely rapidly and people were experimenting with all kinds of things. Since then the language has evolved immensely and stabilized. Some of the features that we used have still not made it in (e.g. decorators) while some that we chose not to adopt widely are now commonplace (e.g. `...` object spread syntax).

This is also the right time to move to TypeScript. When we started, there was no consensus on what the right choice for scalable javascript (e.g. type safety) would be. Since then, TypeScript has emerged as the obvious choice.

Moving to TypeScript means that we will have to continue relying on transpilation, but at least the ecosystem for transpilation is much less convoluted today.

### Development practices and tooling
We first used gulp and browserify, but then moved to webpack to build front-end JS. We kept our old system for SCSS though, so that still relies on gulp, as does server-side transpilation (which is mostly redundant nowadays, but required due to some minor language features that still haven't made it into the standard).

As part of a rebuild we should move entirely to webpack. We should also implement linting to enforce a common code style, and testing where it makes sense.

### Utilize the modern web
The new Organize should be a PWA with reasonable support for installability, offline access, notifications et c. It should be mobile first (although a lot of features will remain more useful on bigger screens).

## A UX case for revamping Organize
Organizers have been using Zetkin for close to four years, and we've learned a lot in that time about what features are actually used and how. 

### Task-driven over data-driven
[...]

Issue #43, opened 2015-08-06
Form validation


richardolsson:
Augment the `Form` and `*Input`/`InputBase` components with support for input validation (required fields, format, minimum length et c).

PR #1144, opened 2020-11-01
Custom Fields in PersonForm

### LINKED ISSUE #1122: Edit custom field values
Add UI to edit custom field values.

One suggestion is to just add the field as regular fields in the `PersonForm`, but handle them differently when submitted.

richardolsson:
This PR adds custom fields to `PersonForm` in both `AddPersonPane` and `EditPersonPane`. This means that values for custom fields can now be set/reset the same way that values for "native" person fields.

Closes #1122

### DIFF
--- a/src/components/forms/PersonForm.jsx
+++ b/src/components/forms/PersonForm.jsx
@@ -1,56 +1,91 @@
 import React from 'react';
+import { connect } from 'react-redux';
 
 import Form from './Form';
+import DateInput from './inputs/DateInput';
 import TextInput from './inputs/TextInput';
 import SelectInput from './inputs/SelectInput';
 
 
 export default class PersonForm extends React.Component {
     render() {
-        var person = this.props.person || {};
-        var genderOptions = {
-            '_': 'forms.person.genderOptions.unknown',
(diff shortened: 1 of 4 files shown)


niklasva82 (approved):
Looks good! One issue is the lack of validation of URLs. PersonField doesn't do client-side validation at this point though so that's something for the future.

richardolsson:
> Looks good! One issue is the lack of validation of URLs. PersonField doesn't do client-side validation at this point though so that's something for the future.

Thanks! I'm not sure what you were referring to here regarding client-side validation. Is this something that we should create a separate issue for?

niklasva82:
There is one, albeit quite old. #43 
Wasn't this an issue at the last code camp? Or was it something related perhaps?

richardolsson:
Ah, of course! I understand now. I was confused by the sheer amount of pull requests and lost track of the context. The form validation is one aspect. It was started at a code camp, but to be honest I don't think we should prioritize it because I believe we can do better with a proper UI framework in a future revamp (#1140).

This particular case could easily be covered by using an `<input type="url">` though, so maybe add an issue for creating a `<URLInput>` component?

PR #1145, opened 2020-11-02
Custom field view columns

richardolsson:
Support custom fields in view columns of type `person_field`.

To test:

1. Navigate to Views
2. Create a view
3. Click "Add Column"
4. From the "Person field" settings, select a custom field

If there are no custom fields, make sure one exists. Either add through the API or reset development data if possible.

image


niklasva82 (approved):
Works for me! Looks good!

PR #1150, opened 2020-11-04
Feature/person field query

niklasva82:
Query on person field

### DIFF
--- a/src/components/filters/FilterTimeFrameSelect.jsx
+++ b/src/components/filters/FilterTimeFrameSelect.jsx
@@ -10,7 +10,7 @@ import IntInput from '../forms/inputs/IntInput';
 export default class FilterTimeFrameSelect extends React.Component {
     static defaultProps = {
         future: true,
-        inlast: false,
+        inlast: true,
     }
 
     constructor(props) {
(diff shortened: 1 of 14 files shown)


src/components/filters/FilterTimeFrameSelect.jsx
```diff
@@ -91,14 +110,22 @@ export default class FilterTimeFrameSelect extends React.Component {
                 after = todayStr;
                 before = today.addDays(30).format('{yyyy}-{MM}-{dd}');
                 break;
+            case 'inlast':
+                days = this.state.days || '30';
+                after = `-${days}d`;
```
richardolsson, line 115: I'm wondering if we should add a `before` here as well. The label implies that matches will be found from the last 30 days, starting 30 days ago until today, but the actual search is for anything after 30 days ago (i.e. also in the future). This is only relevant for campaign participation filter. But maybe it would make sense?

Based on your understanding from having worked with this feature, would that be difficult? What would happen to any old queries that lack a `before` when they're opened in the interface if we add `before`. It's only added after save, right?

richardolsson (commented):
Nice work! I've tried it (and every other filter) with live data and it works as expected!

The code looks fine. I just had one thing that I was thinking about. See comment below, and let me know if you think this is something we should change or if we should leave it be.

src/components/filters/FilterTimeFrameSelect.jsx
niklasva82: You mean like before='now'? No I don't think it would be difficult, but yes, any queries already created would have the old behavior.

richardolsson: Exactly! Do you want to make that change before we merge?

PR #1152, opened 2020-11-04
Confirm CSV export

### LINKED ISSUE #1147: Confirm CSV export
Add a step to confirm CSV export, advising against it for GDPR and feature reasons.

richardolsson:
This PR adds a new pane which is opened when the user clicks to download a view as CSV. The pane contains a message and a button to execute the download.

In the future, we can add options here to configure whether the download should contain all objects or just the visible ones (after filtering).

image

Closes #1147


niklasva82 (approved):
Great! Perhaps this warning could be added to the regular export as well.

richardolsson:
> Great! Perhaps this warning could be added to the regular export as well.

Honestly I foresee us deprecating and removing the regular export. It works so poorly for anything other than small datasets and provides much less control. So lets give that some thought before we put any more effort into it. :)

PR #1153, opened 2020-11-05
Enhancement/before today

niklasva82:
I pushed this to a separate pull request so we can add it later if we want to.

Adds a before: "+1d" to "In last few days" filters in both FilterTimeSelect which is used by most date filters and the CallHistoryFilter. This way only dates up until today are included.


richardolsson:
Supersedes #1150

richardolsson (approved):
Looking good and works as expected. I just fixed a bug which was causing field types not to load if you hadn't already visited a page that preloaded them.

Issue #1156, opened 2020-11-07
Prevent too long input


richardolsson:
No forms restrict length of input values, but the API does. All input fields should have max lengths that match the API (if any). 

I'll come back with the actual maximum lengths. Ask me if I forget. 🙄

kristofferlarberg:
I'll take this

richardolsson:
Fixed by #1159

PR #1159, opened 2020-11-10
Enhancement/textlimit

kristofferlarberg:
Resolves #1159

### DIFF
--- a/src/components/forms/inputs/InputBase.jsx
+++ b/src/components/forms/inputs/InputBase.jsx
@@ -10,7 +10,7 @@ export default class InputBase extends React.Component {
         labelMsg: React.PropTypes.string,
         initialValue: React.PropTypes.string,
         className: React.PropTypes.string,
-        maxLength: React.PropTypes.int
+        maxLength: React.PropTypes.number
     };
 
     constructor(props) {
--- a/src/components/forms/inputs/InputBase.jsx
+++ b/src/components/forms/inputs/InputBase.jsx
@@ -9,7 +9,8 @@ export default class InputBase extends React.Component {
         label: React.PropTypes.string,
         labelMsg: React.PropTypes.string,
         initialValue: React.PropTypes.string,
-        className: React.PropTypes.string
+        className: React.PropTypes.string,
+        maxLength: React.PropTypes.int
     };
 
     constructor(props) {
--- a/src/components/forms/inputs/TextInput.jsx
+++ b/src/components/forms/inputs/TextInput.jsx
@@ -5,16 +5,23 @@ import InputBase from './InputBase';
 
 @injectIntl
 export default class TextInput extends InputBase {
+
+    constructor(props) {
+    super(props);
+    }
+
     renderInput() {
-        let placeholder;
+        let placeholder; 
 
         if(this.props.placeholder) {
             placeholder = this.props.intl.formatMessage(
                 { id: this.props.placeholder });
         }
 
+        const maxLength = this.props.maxLength ? this.props.maxLength : 0
+
         return (
-            <input type="text" value={ this.props.value }
+            <input type="text" maxLength={ maxLength } value={ this.props.value }
                 placeholder={ placeholder }
                 onChange={ this.onChange.bind(this) }/>
         );
--- a/src/components/forms/inputs/TextInput.jsx
+++ b/src/components/forms/inputs/TextInput.jsx
@@ -6,10 +6,6 @@ import InputBase from './InputBase';
 @injectIntl
 export default class TextInput extends InputBase {
 
-    constructor(props) {
-    super(props);
-    }
-
     renderInput() {
         let placeholder; 
 
--- a/src/components/forms/inputs/TextInput.jsx
+++ b/src/components/forms/inputs/TextInput.jsx
@@ -18,7 +18,7 @@ export default class TextInput extends InputBase {
                 { id: this.props.placeholder });
         }
 
-        const maxLength = this.props.maxLength ? this.props.maxLength : 0
+        const maxLength = this.props.maxLength ? this.props.maxLength : null
 
         return (
             <input type="text" maxLength={ maxLength } value={ this.props.value }
--- a/src/components/forms/inputs/TextInput.jsx
+++ b/src/components/forms/inputs/TextInput.jsx
@@ -14,10 +14,8 @@ export default class TextInput extends InputBase {
                 { id: this.props.placeholder });
         }
 
-        const maxLength = this.props.maxLength ? this.props.maxLength : null
-
         return (
-            <input type="text" maxLength={ maxLength } value={ this.props.value }
+            <input type="text" maxLength={ this.props.maxLength } value={ this.props.value }
                 placeholder={ placeholder }
                 onChange={ this.onChange.bind(this) }/>
         );
--- a/src/components/filters/SurveyOptionFilter.jsx
+++ b/src/components/filters/SurveyOptionFilter.jsx
@@ -6,7 +6,6 @@ import FilterBase from './FilterBase';
 import Form from '../forms/Form';
 import RelSelectInput from '../forms/inputs/RelSelectInput';
 import SelectInput from '../forms/inputs/SelectInput';
-import TextInput from '../forms/inputs/TextInput';
 import { retrieveSurveys, retrieveSurvey } from '../../actions/survey';
 
 
--- a/src/components/forms/ActionForm.jsx
+++ b/src/components/forms/ActionForm.jsx
@@ -68,7 +68,7 @@ export default class ActionForm extends React.Component {
                     onEdit={ this.props.onEditActivity }
                     initialValue={ action.activity.id }/>
                 <TextInput labelMsg="forms.action.title" name="title"
-                    initialValue={ action.title }/>
+                    initialValue={ action.title } maxLength={ 300 }/>
                 <TextArea labelMsg="forms.action.info" name="info_text"
                     initialValue={ action.info_text }/>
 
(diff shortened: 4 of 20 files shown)


richardolsson:
Nice work! Will review shortly. I updated the pull request description to use the correct phrasing (removed the word "issue") for GitHub automation to work.

src/components/forms/inputs/TextInput.jsx
```diff
@@ -5,16 +5,23 @@ import InputBase from './InputBase';
 
 @injectIntl
 export default class TextInput extends InputBase {
+
+    constructor(props) {
+    super(props);
+    }
```
niklasva82, line 11: This is a remnant from your earliest attempt I believe?

niklasva82 (changes requested):

src/components/forms/inputs/TextInput.jsx
```diff
@@ -5,16 +5,23 @@ import InputBase from './InputBase';
...
     renderInput() {
-        let placeholder;
+        let placeholder; 
 
         if(this.props.placeholder) {
             placeholder = this.props.intl.formatMessage(
                 { id: this.props.placeholder });
         }
 
+        const maxLength = this.props.maxLength ? this.props.maxLength : 0
```
niklasva82, line 21: maxLength 0 means no characters allowed at all. Setting maxLength to null or undefined should do the trick.

richardolsson: Have you actually tried this @niklasva82? It seemed to work for me.

niklasva82:
I tried it on a test html file and had the limit of zero characters
imposed. I suppose react could process it differently though. Probably best
to change to null for clarity in any case.

Den tis 10 nov. 2020 12:38Richard Olsson <notifications@github.com> skrev:

src/components/forms/inputs/TextInput.jsx
kristofferlarberg: I tried adding a state prop, but the constructor and the rest was already there as far as I remember. Should I remove it anyway?

richardolsson: Thanks for removing it!

The constructor was added in your commit b349b3d1c3c56d5651f5bcde45236b63455b4e43. You can click to see the diff.  It's so easy to forget what was done between two commits, especially when doing trial-and-error of different approaches.

I recommend always using `git add -p` to iterate over all changes and explicitly decide whether they should be part of your commit or not.

It's at least a good idea to inspect what you're about to commit using `git diff` and `git diff --cached`, to make sure that you're not committing any changes that are not required for the work that you're doing.

Maybe your editor (VS Code?) has other good ways of indicating this. I just like to be as close as possible to the code, which is why I like the `git` command-line interface. 😄

src/components/forms/inputs/TextInput.jsx
```diff
@@ -5,16 +5,19 @@ import InputBase from './InputBase';
 
 @injectIntl
 export default class TextInput extends InputBase {
+
     renderInput() {
-        let placeholder;
+        let placeholder;
```
olleolleolle, line 10: Extra whitespace, which can be dropped, for a smaller diff.

src/components/forms/inputs/TextInput.jsx
```diff
@@ -5,16 +5,19 @@ import InputBase from './InputBase';
...
         if(this.props.placeholder) {
             placeholder = this.props.intl.formatMessage(
                 { id: this.props.placeholder });
         }
 
+        const maxLength = this.props.maxLength ? this.props.maxLength : null
+
         return (
-            <input type="text" value={ this.props.value }
+            <input type="text" maxLength={ maxLength } value={ this.props.value }
```
olleolleolle, line 20: Would the thing behave differently, with `maxLength={ this.props.maxLength }` here?

olleolleolle (approved):
I read the whole thing, and the property is applied throughout, 👍 

Looks good. Any changes to the component are easy to do, now that the maxLength values are added all over the app. Super!

src/components/forms/inputs/TextInput.jsx
```diff
@@ -18,7 +14,7 @@ export default class TextInput extends InputBase {
                 { id: this.props.placeholder });
         }
 
-        const maxLength = this.props.maxLength ? this.props.maxLength : 0
+        const maxLength = this.props.maxLength ? this.props.maxLength : null
```
niklasva82, line 17: Agree with Olle here, this is superfluous because if this.props.maxLength is not supplied as a prop, this.props.maxLength will be undefined, and in Javascript null == undefined (though nul !== undefined)

niklasva82 (changes requested):
Just one tiny change. Not a big deal but if you can get away with less code, then why not? Otherwise everything looks good!

src/components/forms/inputs/TextInput.jsx
richardolsson: Again I must ask, are you sure that React will treat `null` and `undefined` the same way for the `maxLength` property? If we are certain it will, I agree that passing through `this.props.maxLength` is nicer.

src/components/forms/inputs/TextInput.jsx
kristofferlarberg: Thanks! I'll do that from now on. I just got very used to `git add .` even if I know there are other ways. Sounds good being more aware of which changes are being commited.

src/components/forms/inputs/TextInput.jsx
kristofferlarberg: Just so I understand this correctly. Should I remove the variable with ternary and change `maxLength` to`this.props.maxLength` in return?

src/components/forms/inputs/TextInput.jsx
richardolsson: I would very strongly recommend against ever using `git add .`. 😄 

When you do, it's very easy to accidentally add files that shouldn't be tracked at all, e.g. swap files or other temporary files, any notes or other personal files you may have put in the working directory, and so on.

On option is `git add -N .`, which will tell Git to care about all files, but not stage them for commit. After that, you can do `git add -p` to go over all new files (and other changes) one by one.

src/components/forms/inputs/TextInput.jsx
richardolsson: Yeah, exactly! Verify that it works, and if it does, that's the cleanest solution.

kristofferlarberg: Done!

niklasva82: Yup, that's correct!

niklasva82: > Again I must ask, are you sure that React will treat `null` and `undefined` the same way for the `maxLength` property? If we are certain it will, I agree that passing through `this.props.maxLength` is nicer.

I tried it out in a react sandbox and it worked fine.

src/components/forms/inputs/TextInput.jsx
kristofferlarberg: That is good to know! Thanks!

src/components/forms/inputs/InputBase.jsx
```diff
@@ -9,7 +9,8 @@ export default class InputBase extends React.Component {
         label: React.PropTypes.string,
         labelMsg: React.PropTypes.string,
         initialValue: React.PropTypes.string,
-        className: React.PropTypes.string
+        className: React.PropTypes.string,
+        maxLength: React.PropTypes.int
```
niklasva82, line 13: Great job! One final thing I noticed is that I get a warning when navigating to Actions.
```
Warning: Failed prop type: TextInput: prop type `maxLength` is invalid; it must be a function, usually from React.PropTypes.
    in TextInput (created by InjectIntl(TextInput))
    in InjectIntl(TextInput) (created by ActionForm)
    in li (created by Form)
    in ul (created by Form)
    in form (created by Form)
    in Form (created by ActionForm)
    in ActionForm (created by EditActionPane)
    in div (created by EditActionPane)
    in div (created by EditActionPane)
    in EditActionPane (created by InjectIntl(EditActionPane))
    in InjectIntl(EditActionPane) (created by Connect(InjectIntl(EditActionPane)))
    in Connect(InjectIntl(EditActionPane)) (created by Section)
    in div (created by Section)
    in div (created by Section)
    in Section (created by InjectIntl(Section))
    in InjectIntl(Section) (created by App)
    in div (created by App)
    in div (created by App)
    in body (created by App)
    in html (created by App)
    in App (created by DragDropContext(App))
    in DragDropContext(App) (created by Connect(DragDropContext(App)))
    in Connect(DragDropContext(App))
    in IntlProvider (created by IntlReduxProvider)
    in Provider (created by IntlReduxProvider)
    in IntlReduxProvider (created by Connect(IntlReduxProvider))
    in Connect(IntlReduxProvider) warning.js:33:14
```

Javascript does not have an integer type (unfortunately imo), you have to use number here for the proptype to be properly interpreted.

niklasva82 (changes requested):

src/components/forms/inputs/InputBase.jsx
kristofferlarberg: Ah! This should be fixed now.

niklasva82: Great! It seems there are other issues related to the propTypes, not related to this pull request. I will create a new issue for those.

niklasva82 (approved):

richardolsson:
Great work!

It seems the reference to the issue in the PR description was wrong. It referenced itself. The correct issue that this PR resolves is #1156.

PR #1160, opened 2020-11-12
Enhancement/editable textlimit

### LINKED ISSUE #1158: Handle limitations in EditableText

kristofferlarberg:
Adds maxLength prop to EditableText. Resolves #1158


kristofferlarberg:
#1160

richardolsson:
Almost there! As discussed on Zoom, please create a separate pull request which fixes the issues of #1158, taking care to avoid unnecessary commits and changes.

PR #1163, opened 2020-11-17
Enhancement/editable text limit

### LINKED ISSUE #1158: Handle limitations in EditableText

kristofferlarberg:
Resolves #1158. Adds limitation to EditableText via maxLength prop as well as limits possibility to make newlines via onKeyDown prop.

There is an issue with contenteditable adding a newline at the end of an entered text. This makes it possible to make one newline upon making changes to the previously added text. This is related to the fact that contenteditable accepts formatting onpaste (which is cleared onBlur).

### DIFF
--- a/src/components/misc/EditableText.jsx
+++ b/src/components/misc/EditableText.jsx
@@ -44,6 +44,21 @@ export default class EditableText extends React.Component {
                     editing: false,
                 });
             },
+            onInput: ev => {
+                if (this.props.maxLength && ev.target.innerText.length <= this.props.maxLength) {
+                    this.setState({
+                        input: ev.target.innerText
+                    });
+                }
+                else if (this.props.maxLength && ev.target.innerText.length > this.props.maxLength) {
+                    ev.target.innerText = this.state.input
+                }
+                else {
+                    this.setState({
+                        input: ''
+                    });
+                };
+            },
             dangerouslySetInnerHTML: {
                 __html: this.state.editing? this.props.content : (this.props.content || this.props.placeholder),
             },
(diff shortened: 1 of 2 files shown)


niklasva82 (changes requested):

niklasva82 (changes requested):
Everything appears to work now! Great work!

There is a bit of a strange setting of state.text in onFocus though, which is not used anywhere else. This appears to do nothing and should probably be removed.

kristofferlarberg:
Great! Yes, I've been wondering about those lines. I'll remove them.

niklasva82 (approved):
Great work! You need to make this a regular PR rather than a draft PR for it to be merged!

richardolsson:
I would actually like to request that you create a new pull request which is more to the point, like we discussed previously with another PR. Or you could squash together commits into a few succinct commits. There is a lot of experimentation going on in this pull request, and it would be nice to keep the commit history a little bit cleaner post merging.

Are you ok with that @kristofferlarberg?

kristofferlarberg:
Of course, I'll try squashing the commits into two? 
1. onKeyDown maxlength and application of prop 
2. onKeyDown with prevEnter and application of prop

richardolsson:
> Of course, I'll try squashing the commits into two?
> 
> 1. onKeyDown maxlength and application of prop
> 2. onKeyDown with prevEnter and application of prop
> 
> Edit: Will make a new pull request instead.

Sounds good! The property for single-/multi-line should really be called something like just that. Probably `multiline={ true}`. Two reasons for this:

* In general it's a good idea to avoid "negative" properties
* The `preventEnter` name is an implementation detail. What happens if you paste several lines, i.e. insert newlines without pressing "enter". That should also be prevented.

Looking forward to a separate PR! Sorry for the hassle. 🙄

PR #1164, opened 2020-11-22
BUGFIX: Truncate view column title to a max of 80 characters

niklasva82:
Creating a person view column from a survey question results in the column title having the same title as the survey question. But the max length of a survey question is over 200 characters whereas the max length of a column title is 80 characters.

This fix truncates column titles in the createPersonViewColumn action before it is sent to the server.

### DIFF
--- a/src/actions/personView.js
+++ b/src/actions/personView.js
@@ -57,6 +57,10 @@ export function createPersonViewColumn(viewId, data) {
     return ({ dispatch, getState, z }) => {
         const orgId = getState().org.activeId;
 
+        if(data.title.length > 80) {
+          data.title = data.title.substring(0,77) + '...';
+        }
+
         dispatch({
             type: types.CREATE_PERSON_VIEW_COLUMN,
             meta: { viewId },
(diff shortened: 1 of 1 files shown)


src/actions/personView.js
```diff
@@ -57,6 +57,10 @@ export function createPersonViewColumn(viewId, data) {
     return ({ dispatch, getState, z }) => {
         const orgId = getState().org.activeId;
 
+        if(data.title.length > 80) {
+          data.title = data.title.substring(0,77) + '...';
```
richardolsson, line 61: Deviating indentation on this line.

richardolsson (changes requested):
This is the simplest possible fix, but should work fine. I just had issue with the indentation which for some reason deviates from the rest of the file (and project standard).

src/actions/personView.js
niklasva82: Ah yeah it's because I tried a new distribution of vim which had 2 character-indentation as default :P fixed.

richardolsson (approved):
This PR almost escaped, but looks good so I will merge and include in R24.

Issue #1157, opened 2020-11-09
Font-size increased when pasting text in person view title


niklasva82:
To reproduce:
Select title in PersonView and copy, then paste over and over.

Expected result: Font size in TextInput remains constant
Actual result: Font size in TextInput increases for every time user pastes text.

PR #1166, opened 2020-11-24
Enhancement/editabletext limits

### LINKED ISSUE #1158: Handle limitations in EditableText

kristofferlarberg:
Resolves #1158. Adds possibility for text length limitation as well as prevents newlines in EditableText.

### DIFF
--- a/src/components/sections/people/PersonViewsPane.jsx
+++ b/src/components/sections/people/PersonViewsPane.jsx
@@ -174,6 +174,7 @@ export default class PersonViewsPane extends RootPaneBase {
                             content={ viewItem.data.title }
                             onChange={ this.onChange.bind(this, 'title') }
                             placeholder={ this.props.intl.formatMessage({ id: 'panes.personViews.placeholders.title' }) }
+                            preventEnter={ true }
                             />
                         <EditableText tagName="p" key="description"
                             content={ viewItem.data.description }
--- a/src/components/sections/people/PersonViewsPane.jsx
+++ b/src/components/sections/people/PersonViewsPane.jsx
@@ -175,6 +175,7 @@ export default class PersonViewsPane extends RootPaneBase {
                             onChange={ this.onChange.bind(this, 'title') }
                             placeholder={ this.props.intl.formatMessage({ id: 'panes.personViews.placeholders.title' }) }
                             preventEnter={ true }
+                            maxLength={ 80 }
                             />
                         <EditableText tagName="p" key="description"
                             content={ viewItem.data.description }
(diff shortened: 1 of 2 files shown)


src/components/sections/people/PersonViewsPane.jsx
```diff
@@ -174,6 +174,8 @@ export default class PersonViewsPane extends RootPaneBase {
                             content={ viewItem.data.title }
                             onChange={ this.onChange.bind(this, 'title') }
                             placeholder={ this.props.intl.formatMessage({ id: 'panes.personViews.placeholders.title' }) }
+                            preventEnter={ true }
```
richardolsson, line 177: I think this property should be renamed to `multiline` or something like that (which is the inverse of what it does now, so logic will need to be slightly altered as well).

richardolsson (changes requested):
Looking good! I'll let @niklasva82 review the main logic. I had just one comment on the naming of a property.

I'm also curious what happens when you paste text which contains newline characters. Will it allow that when "preventing" enter?

src/components/sections/people/PersonViewsPane.jsx
kristofferlarberg: Right, so basically the prop will be multiline={false}?

kristofferlarberg:
> Looking good! I'll let @niklasva82 review the main logic. I had just one comment on the naming of a property.
> 
> I'm also curious what happens when you paste text which contains newline characters. Will it allow that when "preventing" enter?

Hmm, yes, at the moment all newlines are are removed onBlur, but this seems to be default? The const I added replaces newline with "" in order to make the field entirely empty when no text and hence show the placeholder text. Makes sense?

niklasva82:
> > Looking good! I'll let @niklasva82 review the main logic. I had just one comment on the naming of a property.
> > I'm also curious what happens when you paste text which contains newline characters. Will it allow that when "preventing" enter?
> 
> Hmm, yes, at the moment all newlines are removed onBlur, but this seems to be default? The const I added replaces newline with "" in order to make the field entirely empty when no text and hence show the placeholder text. Makes sense?

This works and I think delving into the clipboard API is probably overkill at this point. At some point we should fix #1157 which is related to this and could be fixed by removing formatting from incoming pasted data, but I think this is outside the scope of this issue.

I do think the preventEnter prop should be renamed to multiline and inverted as @richardolsson suggested though before we merge this, it makes for a more easy to understand interface for this component.

kristofferlarberg:
> I do think the preventEnter prop should be renamed to multiline and inverted as @richardolsson suggested though before we merge this, it makes for a more easy to understand interface for this component.

This is changed now and works!

richardolsson:
Will you review again @niklasva82?

niklasva82 (approved):
Great work!

There is a bug where if you feed a newline and then start typing, you can type outside the EditableText window. But I think this was there before this change and I will file a bug report if there isn't one already.

richardolsson (approved):

PR #1167, opened 2020-12-01
Feature/organization import

### LINKED ISSUE #1063: Map to sub-organization on import
Sub-organizations and organizational sharing is currently being implemented in the Zetkin API. This will make it possible for organizations to import people into it's sub-organizations.

Implement a mapping interface for this, i.e. to map text content in a column to a particular sub-organization, similarly to how tags are mapped today.

niklasva82:
Organization import
Resolves #1063 

Note that currently, fetching recursive organizations is done by simply adding ?recursive in the subOrg action. This works for now, but for future uses of this flag (especially on paginated resources), I added an issue at zetkin/zetkin.js#32 for discussion regarding this.


src/components/misc/importer/settings/PersonOrganizationColumnSettings.jsx
```diff
@@ -0,0 +1,78 @@
...
+                        <li key={ value || 0 }
+                            className="PersonOrganizationColumnSettings-mapping">
+                            <Msg tagName="h3" id={ labelMsg }
+                                values={{ value }} />
+                            <SelectInput
+                                value={ mapping.org }
+                                name={ value }
+                                options={ this.orgOptions }
+                                optionLabelsAreMessages={ false }
+                                onValueChange={ this.onOrganizationChange.bind(this) }
```
richardolsson, line 52: There needs to be a null option here, so that we're not forced to assign all rows to an organization.

Alternatively, maybe we could rename the one that reflects the current organization so that it indicates that it's synonymous to picking null?

See screenshot below (which also shows the effect of the misspelled `emptyLabel` string mentioned elsewhere.)

image

richardolsson (changes requested):
Great work on this! I've finally looked through the entire diff, and also tried it in action. It looks mostly good, but I do have some comments below.

src/components/misc/importer/settings/PersonOrganizationColumnSettings.jsx
niklasva82: Yeah that's pretty much what I intended, for the current organization to be the "null"-option.

src/actions/importer.js
```diff
@@ -158,6 +180,26 @@ export function executeImport(tableId) {
             columns[genderIdx].config = { field: 'gender' }
         }
 
+        // Find the organization field if it exists
+        let orgIdx = columns.findIndex(i => i.type == 'organization');
```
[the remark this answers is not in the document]
niklasva82: It did, fixed now but this requires a fix in import service, linked into this issue.

richardolsson:
I've looked at the code which looks good! Will try the full integration in action next.

richardolsson (approved):
Confirmed the full integration to be working!

PR #1190, opened 2020-12-02
Custom field overrides during merge

### LINKED ISSUE #1162: Merging with custom fields
When there are conflicting custom fields, the user is not currently prompted to pick the value for the overrides.

richardolsson:
This PR adds capabilities to select which custom field values to keep after a person merge. To test it, create some duplicates with conflicting custom field values, and try merging. While merging, try selecting different values. Also try merging when there are no custom fields on a person, or when only one of the duplicates has values assigned.

Closes #1162


niklasva82 (approved):
Looks good and works for me!

Issue #1118, opened 2020-10-16
Sticky PersonViewTableHead and avatars


richardolsson:
As views grow bigger and extend outside of the viewport the user needs to scroll vertically and horizontally, and it's very easy to lose track of both the columns and rows.

The header should become sticky at the top of the view, and the avatars should become sticky on the left-hand side.

richardolsson:
Relevant to #1155 and #1174

richardolsson:
Partly resolved by #1194. The sticky Avatars aspect of this enhancement is difficult to resolve in a performant way. Postponing that until further notice.

PR #1194, opened 2020-12-03
View scrolling improvements

### LINKED ISSUE #1155: Horizontal guides in view
Views would be much easier to read if there was some sort of horizontal visual guide. Either lines between the rows, or alternating row colors.

### LINKED ISSUE #1174: Fix view scrolling to pane
The horizontal scrollbar (if necessary) appears at the very bottom of the view pane, which is vertically scrolled separately. This means that a user has to scroll all the way down to find the horizontal scrollbar.

Instead, the scrolling should be fixed to the pane rectangle.

richardolsson:
This PR makes a number of improvements to the experience when working with large views that require scrolling.

* Fixate view table to browser viewport, so that horizontal and vertical scrollbars are visible. Fixes #1174 
* Keep table header sticky at top. Fixes half of #1118 (sticky avatars proved very complex and must be postponed)
* Add cell outlines Fixes #1155 

See screenshots below for examples of what this looks like in default state, when scrolled a little bit in both directions, and the new location of the "Add" widget at the bottom.

image
image
image

### DIFF
--- a/src/components/misc/personViews/PersonViewAddRow.jsx
+++ b/src/components/misc/personViews/PersonViewAddRow.jsx
@@ -1,0 +1,23 @@
+import React from 'react';
+import { connect } from 'react-redux';
+
+import RelSelectInput from '../../forms/inputs/RelSelectInput';
+import PersonSelectWidget from '../PersonSelectWidget';
+
+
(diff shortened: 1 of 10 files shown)


niklasva82 (approved):
Looks great and works as advertised!

PR #1196, opened 2020-12-04
Filtering and paging of Join Forms

### LINKED ISSUE #1125: Join submission pagination and filtering
Implement pagination and filtering of join submissions (based on form and state).

Related to #1101.

niklasva82:
Resolves #1125

### DIFF
--- a/src/components/sections/people/JoiningPane.scss
+++ b/src/components/sections/people/JoiningPane.scss
@@ -11,4 +11,15 @@
     .ViewSwitch-state-forms:before {
         @include icon($fa-var-cog);
     }
+
+    .RootPaneBase-toolbar-column-reverse {
+        display: flex;
+        flex-direction: column-reverse;
+        margin-left: -1em;
+
+        .RootPaneBase-filterButton {
+            margin-top: 0.5em;
+            margin-left: 0.8em;
+        }
+    }
 }
--- a/src/components/sections/people/JoiningPane.scss
+++ b/src/components/sections/people/JoiningPane.scss
@@ -12,6 +12,26 @@
         @include icon($fa-var-cog);
     }
 
+    .RootPaneBase-toolbar {
+        display: flex;
+
+        .viewSwitch {
+            margin-right: 1em;
+            margin-left: 1em;
+            order: 0;
+        }
+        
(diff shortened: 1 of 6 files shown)


richardolsson:
Note to self while reviewing (and perhaps to you @niklasva82 in case you didn't think of this already).

Deleting a person should update the store and remove all join submissions matching that person, and ideally close any related JoinSubmissionPanes. Steps to test:

1. Open a join submission pane from the list
2. Approve if not approved already
3. Click the person to open a person pane
4. Edit person date
5. Delete person

Expected results: The join submission should disappear from the list, and all three panes should close.

richardolsson (changes requested):
Code looks good and seems to work as expected. The only thing is that I'd like the filter button to move. Does that make sense?

image

niklasva82:
> Code looks good and seems to work as expected. The only thing is that I'd like the filter button to move. Does that make sense?
> 
> image

Sure! That actually 

> Code looks good and seems to work as expected. The only thing is that I'd like the filter button to move. Does that make sense?
> 
> image

Done, that also removed the need for changes in RootPaneBase :)

richardolsson:
> Done, that also removed the need for changes in RootPaneBase :)

Ah, I see! Maybe remove the CSS also then?

src/components/sections/people/JoiningPane.scss
```diff
@@ -11,4 +11,35 @@
...
+    .RootPaneBase-toolbar-column-reverse {
+        display: flex;
+        flex-direction: column-reverse;
+        margin-left: -1em;
+
+        .RootPaneBase-filterButton {
+            margin-top: 0.5em;
+            margin-left: 0.8em;
+        }
+    }
```
richardolsson, line 44: Is this still needed? Or only part of it?

richardolsson (changes requested):

src/components/sections/people/JoiningPane.scss
niklasva82: Yeah the toolbar-column-reverse is no longer needed. Fixed.

richardolsson (approved):
Nice! We've arrived!

Issue #1197, opened 2020-12-04
Generalize native person fields


richardolsson:
Native person fields appear all over the place, but the translations and some other basic utilities are repeated over and over. I'm thinking about:

* Localizations of the native field labels
* List of `NATIVE_FIELDS`
* Logic to display fields belonging to a person, regardless of native or custom

All of these should be generalized and reused across various features like `PersonPane`, `JoinSubmissionPane`, views et c.

richardolsson:
"Native fields" refer to those person fields which are not custom and hence exist in all organizations. Native fields include `first_name`, `last_name`, `email`, `phone` etc.

PR #1198, opened 2020-12-04
More information in JoinSubmissionPane

### LINKED ISSUE #1169: All form fields in join submission pane
Include all JoinForm fields in the `JoinSubmissionPane`, not just native fields.

richardolsson:
This PR makes the following changes to JoinSubmissionPane:

* Show all form fields while submission is pending. Fixes #1169
* Move meta-data section to top of pane and add icons
* Tweak styling

image

image

### DIFF
--- a/src/components/panes/JoinSubmissionPane.jsx
+++ b/src/components/panes/JoinSubmissionPane.jsx
@@ -120,8 +120,13 @@ export default class JoinSubmissionPane extends PaneBase {
                     <ul className="JoinSubmissionPane-personData">
                     {this.props.formItem.data.fields.map(fieldName => {
                         const fieldItem = this.props.fieldTypes.items.find(item => item.data.slug == fieldName);
-                        const label = fieldItem? fieldItem.data.title : this.props.intl.formatMessage({ id: `misc.fields.${fieldName}` });
 
+                        // Exclude JSON fields
+                        if (fieldItem && fieldItem.data.type == 'json') {
+                            return null;
+                        }
+
+                        const label = fieldItem? fieldItem.data.title : this.props.intl.formatMessage({ id: `misc.fields.${fieldName}` });
                         const value = person[fieldName];
 
                         return (
--- a/src/components/panes/JoinSubmissionPane.jsx
+++ b/src/components/panes/JoinSubmissionPane.jsx
@@ -35,6 +35,9 @@ const mapStateToProps = (state, props) => {
             subItem: subItem,
         }
     }
+    else {
+        return {};
+    }
 };
 
 @connect(mapStateToProps)
--- a/src/components/panes/JoinSubmissionPane.jsx
+++ b/src/components/panes/JoinSubmissionPane.jsx
@@ -48,13 +48,13 @@ export default class JoinSubmissionPane extends PaneBase {
         this.props.dispatch(retrieveFieldTypesForOrganization());
 
         if (this.props.subItem && this.props.subItem.data && !this.props.formItem) {
-            this.props.dispatch(retrieveJoinForm(nextProps.subItem.data.form.id));
+            this.props.dispatch(retrieveJoinForm(this.props.subItem.data.form.id));
         }
     }
 
     componentDidUpdate() {
         if (this.props.subItem && this.props.subItem.data && !this.props.formItem) {
-            this.props.dispatch(retrieveJoinForm(nextProps.subItem.data.form.id));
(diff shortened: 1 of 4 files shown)


richardolsson:
The PR also introduces a "new" set of translations in `misc.fields`, which includes translations of native fields. This is not used anywhere else yet, and there are also some other DRY issues with fields that I have created a separate issue for, see #1197.

src/components/panes/JoinSubmissionPane.jsx
```diff
@@ -34,6 +45,17 @@ export default class JoinSubmissionPane extends PaneBase {
 
         const subId = this.getParam(0);
         this.props.dispatch(retrieveJoinSubmission(subId));
+        this.props.dispatch(retrieveFieldTypesForOrganization());
+
+        if (this.props.subItem && this.props.subItem.data && !this.props.formItem) {
+            this.props.dispatch(retrieveJoinForm(nextProps.subItem.data.form.id));
```
niklasva82, line 51: Where does nextProps come from?

src/components/panes/JoinSubmissionPane.jsx
```diff
@@ -34,6 +45,17 @@ export default class JoinSubmissionPane extends PaneBase {
...
+        }
+    }
+
+    componentDidUpdate() {
+        if (this.props.subItem && this.props.subItem.data && !this.props.formItem) {
+            this.props.dispatch(retrieveJoinForm(nextProps.subItem.data.form.id));
```
niklasva82, line 57: Same here

niklasva82 (changes requested):

richardolsson:
Fixed the `nextProps` issue. Please review again! :)

src/components/panes/JoinSubmissionPane.jsx
```diff
@@ -100,14 +114,33 @@ export default class JoinSubmissionPane extends PaneBase {
...
+                    <ul className="JoinSubmissionPane-personData">
+                    {this.props.formItem.data.fields.map(fieldName => {
+                        const fieldItem = this.props.fieldTypes.items.find(item => item.data.slug == fieldName);
+
+                        // Exclude JSON fields
+                        if (fieldItem && fieldItem.data.type == 'json') {
+                            return null;
+                        }
+
+                        const label = fieldItem? fieldItem.data.title : this.props.intl.formatMessage({ id: `misc.fields.${fieldName}` });
```
niklasva82, line 129: Before custom fields are loaded, this results in formatMessage() running on misc.fields.{custom field}, which results in an error being logged to the console. e.g. `[React Intl] Cannot format message: "misc.fields.extra_text", using message id as fallback.`

src/components/panes/JoinSubmissionPane.jsx
```diff
@@ -12,19 +12,30 @@ import Button from '../misc/Button';
...
 
-const mapStateToProps = (state, props) => ({
-    subItem: getListItemById(state.joinForms.submissionList,
-        props.paneData.params[0]),
-});
+const mapStateToProps = (state, props) => {
+    const subItem = getListItemById(state.joinForms.submissionList,
+        props.paneData.params[0]);
+
+    if (subItem) {
```
niklasva82, line 30: If not subItem (i.e. when submissionList is not yet loaded), nothing is returned here and this is logged to the console:
`mapStateToProps() in Connect(InjectIntl(JoinSubmissionPane)) must return a plain object. Instead received undefined. warning.js:14:12`

niklasva82 (changes requested):
A couple of issues related to loading of resources. It still works but errors are logged to the console due to this.

src/components/panes/JoinSubmissionPane.jsx
richardolsson: Will fix this when I get a chance.

src/components/panes/JoinSubmissionPane.jsx
richardolsson: This is "by design". The error is only printed in development mode, and is just a warning to indicate that something wasn't translated correctly, which is not a problem. I'll consider checking against `NATIVE_FIELDS` but I don't want to complicate things unnecessarily.

richardolsson:
I've pushed fixes to the issues you found now @niklasva82. Please re-review again. :)

niklasva82 (approved):
Great! Everything seems to work fine now and no errors. Code looks good too.

PR #1199, opened 2020-12-04
Handle survey_response cells when searching in view

### LINKED ISSUE #1142: Filter in view based on survey responses
The filter widget in Views does not search in survey responses.

richardolsson:
This PR handles cells in `survey_response` columns properly when searching in a view. These cells include an array of objects, each object representing a response to the given survey question. The logic merges the array into a single string and searches it the same way that regular search in text-based cells work.

Fixes #1142

### DIFF
--- a/src/components/misc/personViews/PersonViewTable.jsx
+++ b/src/components/misc/personViews/PersonViewTable.jsx
@@ -79,9 +79,20 @@ export default class PersonViewTable extends React.Component {
                         const searchStr = this.state.searchStr.toLowerCase();
 
                         visibleRows = visibleRows.filter(item => {
-                            return item.data.content.some(cell => {
-                                if (cell && cell.toLowerCase) {
-                                    return cell.toLowerCase().indexOf(searchStr) >= 0;
+                            return item.data.content.some((cell, idx) => {
+                                const col = colList.items[idx].data;
+
+                                let text = null;
+
+                                if (cell && col.type == 'survey_response') {
+                                    text = cell.map(r => r.text).join('\n');
+                                }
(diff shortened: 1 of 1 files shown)


niklasva82:
I'm not sure I understand how this PR changes things. It concacenates an entire survey and searches through the entire thing is my understanding, but how do I select an enitre survey? When I try, this is what happens.
Peek 2020-12-04 14-48

richardolsson:
You are not adding any columns (because you are not picking a question). The pull request adds feature to _search_. You already have to `survey_response` columns in this view. Try searching in the View (just next to the download button) for "this is", which will find Angela due to her survey response that you've added to the view.

richardolsson:
> It concacenates an entire survey and searches through the entire thing is my understanding

That's not correct. The `survey_response` column returns an array of responses to a single question, in case that question was answered many times be a single person. In most cases it will have 0 or 1 entry, but in some cases (whey a person answers the same survey more than once) it will have more than 1 entry.

The old logic of searching just understood text cells, but this is a cell which contains an array. The new code joins the array entries within a single cell into a single string of text, and searches that.

niklasva82:
Ah okay now it makes sense! It was the array that I was confused about.

niklasva82 (approved):
Great! Works and code looks good!

richardolsson:
> Ah okay now it makes sense! It was the array that I was confused about.

Cool! I think what you've found (adding a column without finishing all the options) is a bug though. I will log it separately!

richardolsson:
Logged the bug you found as #1200.

PR #1202, opened 2020-12-07
Fix minor styling issues for views in some browsers

richardolsson:
This PR fixes some misalignment issues between table header and rows, especially in Firefox, but also in some cases on Chrome. Related to view scrolling, so make sure to test with a wide and tall view.


niklasva82 (approved):
Looks good! Tested with Firefox.

PR #1204, opened 2020-12-10
some filter for person_tags

niklasva82:
Implements some filter for tags for specifying minimum number of tags in search.


richardolsson (approved):
LGTM!

Issue #1161, opened 2020-11-12
Order alphabetically


richardolsson:
A bunch of lists and select boxes are sorted by the default order from the API, which is not very user friendly.

Lets work together to find them. Here are some that I can think of:

- [x] Smart Search select in `PersonViewsPane`
- [x] Tag select in `AddViewColumnPane`
- [x] Smart Search select in `AddViewColumnPane`
- [x] Tag list in Manage section
- [x] Smart Search list in Manage section
- [x] View items in Views section

richardolsson:
Two items fixed by #1207.

olleolleolle:
#1230 should cover the rest of this list.

richardolsson:
Fixed by #1230

PR #1207, opened 2020-12-11
Sort tags alphabetically in AddViewColumnPane

richardolsson:
Relevant to #1161


niklasva82 (approved):
Looks good! Is there a reason the smart search SelectInput in the same pane did not get the same treatment?

richardolsson:
Good point @niklasva82. How about this? 😄

Issue #1200, opened 2020-12-04
Bug when creating view column


richardolsson:
Steps to reproduce

1. Create a new view
2. Click "Add column"
3. Pick "Survey Response"
4. Click the blue confirm button at the bottom

Expected results: Error message (because you picked no survey)
Actual results: Pane closes, a column of other type is created

This could be fixed by hiding the confirm button until the settings have been properly configured.

sam-mcmurray:
Sure ill take this

PR #1215, opened 2020-12-19
Bug/entrylevel/bug when creating view column

sam-mcmurray:


richardolsson:
This is intended as a fix to #1200

PR #1223, opened 2021-01-21
PersonViewAddRow visible before rows are populated

### LINKED ISSUE #1218: Can't add single person to view
Since #1194, there is no longer an add widget in the view interface unless you first add rows to the view. Steps to reproduce:

1. Create a new blank view
2. Open the empty view (happens automatically when creating it)

Expected results: An interface with an add widget to add a specific person
Actual results: No such interface

niklasva82:
Resolves #1218


richardolsson (approved):
LGTM! 🚀

PR #1224, opened 2021-01-22
Feature/delete view

### LINKED ISSUE #1115: Delete view
Add option to delete a person view. Consider combining this along with other things like settings (e.v. #1109) and view export (#1108) in a settings/misc section or menu.

niklasva82:
Resolves #1115


niklasva82:
delete
I added a delete button next to the download button and created a confirmation pane. What do you think about this design @jensborje?

jensborje:
Great to require a confirmation pane for the action!

One improvement could be the placement. The action is on a higher conceptional layer than the neighbouring inputs in the UI. Ie, both to download and to filter is actions related to the content in the view while the delete button is a general setting/setup action. It's generally better to group these actions on the same level. For deletion, this could be next to saving and naming the view.

If it for some reason need to be where it is right now, one could tweak the margins and separation from the other inputs. Now, it is connected to the download button which signals that these two actions relate and is opposite to each other.

richardolsson:
Great work @niklasva82 and awesome feedback @jensborje!

Maybe it could be made a link (text, not button) in a small font right below the name? This would make the header larger which is a problem, but I propose we solve that by making the entire header collapsable.

I've created a separate issue for the collapsable header. Do you think that's something you could look at next, @niklasva82?

niklasva82:
How's this?
delete

richardolsson:
I think that looks nice! I would just tweak it a little bit, putting it closer to the title (smaller margin) and maybe adding the underline to indicate that it's a link? Also adding back the trashcan icon maybe?

niklasva82:
Alright
screenshot

richardolsson:
Looking good! I'll review this properly next. 😊

src/components/panes/index.js
```diff
@@ -29,6 +29,7 @@ import CampaignPane from './CampaignPane';
 import CanvassAssignmentPane from './CanvassAssignmentPane';
 import ConfirmImportPane from './ConfirmImportPane';
 import ConfirmExportPane from './ConfirmExportPane';
+import ConfirmDeletePane from './ConfirmDeletePane';
```
richardolsson, line 32: This file seems to be missing?

richardolsson (changes requested):
Code looks good but doesn't work because the new pane is missing.

src/components/panes/index.js
niklasva82: Fixed!

richardolsson (commented):
Code looks good and works mostly as expected. I'm wondering if it would be possible to navigate away from the single-view page after deleting?

Currently it displays a spinner after the view is deleted, which implies that the view is loading. I know this is the case in many other places, but not in all. In many places we have resorted to updating the `view` store (note, not `personViews`, but the UI store called `view`) to remove/change panes when an object is deleted.

I'm not sure how difficult that would be in this case, since we're working with a root pane. What do you think @niklasva82?

niklasva82:
Works now!

richardolsson (approved):
All good now! Nice work, loving the collaboration and iterative process. 😊

PR #1227, opened 2021-01-23
Edited getValues to getChangedValues

### LINKED ISSUE #1149: Unify form update method
Forms/edit panes seem to work differently. Some submit the entire form state, while others just submit the fields that changed. Submitting fields that didn't change affects the change-logging and will make it slightly less useful for users once we introduce it into the UI, since the UI will say that a field changed at a time when the user doesn't recognize changing it.

Go over all edit panes/forms and make sure they all use `getChangedValues()` only.

johnschack:
Resolves #1149. Have not been able to test fully

### DIFF
--- a/src/components/panes/EditGroupPane.jsx
+++ b/src/components/panes/EditGroupPane.jsx
@@ -66,7 +66,7 @@ export default class EditGroupPane extends PaneBase {
         ev.preventDefault();
 
         let groupId = this.getParam(0);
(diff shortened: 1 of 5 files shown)


johnschack:
EditTextPane and EditCallerPane does not have onSubmit-function

richardolsson (approved):
Nice work! All seems to be working and the code looks good. 🚀

PR #1229, opened 2021-01-23
Drop unused variables

olleolleolle:
This PR cleans up unused variables.

Paired with @oelrich


richardolsson (approved):
Looking good and still working as expected!

PR #1230, opened 2021-01-23
PersonTagList, PersonQueryList, PersonViewsPane: A-Z order

olleolleolle:
This PR implements A-Z sorting in different places.

We used JS sorting for some things, and options in React Component elsehwere.

See #1161

Paired with @oelrich!


richardolsson (approved):
I looked at the code and tested the functionality. All is well! Great work 🚀

Issue #1210, opened 2020-12-11
Sort on survey_response view column


richardolsson:
Sorting on a `survey_response` column does not yield the expected order. Probably because it's comparing arrays instead of the text that is contained in those arrays (and displayed to the user). See related discussion and fix to view search in #1199.

Steps to reproduce:

1. Create a view
2. Add a `survey_response` column
3. Populate view with enough people that there are at least three rows with values in the `survey_response` column
4. Sort on the `survey_response` column

Expected results: The rows should be sorted alphabetically on the text as it's rendered
Actual results: The rows appear to be sorted randomly (although most likely they are sorted based on arrays being compared)

richardolsson:
Solved by #1231

PR #1231, opened 2021-01-23
Bug/entry level/sort survey response

fannysoderlund:
issue #1210

### DIFF
--- a/src/components/misc/personViews/PersonViewTable.jsx
+++ b/src/components/misc/personViews/PersonViewTable.jsx
@@ -133,13 +133,24 @@ export default class PersonViewTable extends React.Component {
                                 const val1 = row1.data.content[this.state.sortIndex] || '';
 
                                 let x = 0;
-                                
-                                  if (Array.isArray(val0) && typeof(val0[0]) != ‘undefined’ && typeof(val1[0]) != ‘undefined’) {
+
+                                  if (Array.isArray(val0) && typeof(val0[0]) != 'undefined' && typeof(val1[0]) != 'undefined') {
                                      x = val0[0].text.toLowerCase().toString().localeCompare(val1[0].text.toLowerCase().toString());
-                                  }
-                                  else {
+                                  } else {
                                  x = val0.toString().localeCompare(val1.toString());
                               }
+
+                                if(typeof(val0[0]) == 'undefined' && typeof(val1[0]) != 'undefined') {
+                                  return 1;
+                                }
+                                if(typeof(val0[0]) != 'undefined' && typeof(val1[0]) == 'undefined') {
+                                  return -1;
+                                }
+                                if(typeof(val0[0]) == 'undefined' && typeof(val1[0]) == 'undefined') {
+                                  return 0;
+                                }
+
+
                                 if (this.state.sortInverted) {
                                     x *= -1;
                                 }
--- a/src/components/misc/personViews/PersonViewTable.jsx
+++ b/src/components/misc/personViews/PersonViewTable.jsx
@@ -93,7 +93,7 @@ export default class PersonViewTable extends React.Component {
                 } else {
                     placeholder = this.props.placeholder;
                 }
-                
+
                 if (rowList.items && rowList.items.length) {
                     let visibleRows = rowList.items;
 
@@ -128,19 +128,24 @@ export default class PersonViewTable extends React.Component {
 
                     // Sort, if a column is selected for sorting
                     if (this.state.sortIndex !== null) {
-                        visibleRows = visibleRows.concat().sort((row0, row1) => {
-                            const val0 = row0.data.content[this.state.sortIndex] || '';
-                            const val1 = row1.data.content[this.state.sortIndex] || '';
+                            visibleRows = visibleRows.concat().sort((row0, row1) => {
+                                const val0 = row0.data.content[this.state.sortIndex] || '';
+                                const val1 = row1.data.content[this.state.sortIndex] || '';
 
-                            let x = val0.toString().localeCompare(val1.toString());
-
-                            if (this.state.sortInverted) {
-                                x *= -1;
-                            }
-
-                            return x;
-                        });
-                    }
+                                let x = 0;
+                                
+                                  if (Array.isArray(val0) && typeof(val0[0]) != ‘undefined’ && typeof(val1[0]) != ‘undefined’) {
+                                     x = val0[0].text.toLowerCase().toString().localeCompare(val1[0].text.toLowerCase().toString());
+                                  }
+                                  else {
+                                 x = val0.toString().localeCompare(val1.toString());
+                              }
+                                if (this.state.sortInverted) {
+                                    x *= -1;
+                                }
+                                return x;
+                            });
+                        }
 
                     // Store match count for label
                     numMatches = visibleRows.length;
--- a/src/components/misc/personViews/PersonViewTable.jsx
+++ b/src/components/misc/personViews/PersonViewTable.jsx
@@ -134,20 +134,21 @@ export default class PersonViewTable extends React.Component {
 
                                 let x = 0;
 
-                                  if (Array.isArray(val0) && typeof(val0[0]) != 'undefined' && typeof(val1[0]) != 'undefined') {
-                                     x = val0[0].text.toLowerCase().toString().localeCompare(val1[0].text.toLowerCase().toString());
-                                  } else {
-                                 x = val0.toString().localeCompare(val1.toString());
-                              }
+                                if (Array.isArray(val0) && typeof(val0[0]) != 'undefined' && typeof(val1[0]) != 'undefined') {
+                                    x = val0[0].text.toLowerCase().toString().localeCompare(val1[0].text.toLowerCase().toString());
+                                }
+                                else {
+                                    x = val0.toString().localeCompare(val1.toString());
+                                }
 
                                 if(typeof(val0[0]) == 'undefined' && typeof(val1[0]) != 'undefined') {
-                                  return 1;
+                                    return 1;
                                 }
                                 if(typeof(val0[0]) != 'undefined' && typeof(val1[0]) == 'undefined') {
-                                  return -1;
+                                    return -1;
                                 }
                                 if(typeof(val0[0]) == 'undefined' && typeof(val1[0]) == 'undefined') {
-                                  return 0;
+                                    return 0;
                                 }
 
 
--- a/src/components/misc/personViews/PersonViewTable.jsx
+++ b/src/components/misc/personViews/PersonViewTable.jsx
@@ -127,36 +127,36 @@ export default class PersonViewTable extends React.Component {
                     }
 
                     // Sort, if a column is selected for sorting
-            if (this.state.sortIndex !== null) {
-                visibleRows = visibleRows.concat().sort((row0, row1) => {
-                    const val0 = row0.data.content[this.state.sortIndex] || '';
-                    const val1 = row1.data.content[this.state.sortIndex] || '';
+                    if (this.state.sortIndex !== null) {
+                        visibleRows = visibleRows.concat().sort((row0, row1) => {
+                            const val0 = row0.data.content[this.state.sortIndex] || '';
+                            const val1 = row1.data.content[this.state.sortIndex] || '';
 
-                    let x = 0;
+                            let x = 0;
 
-                    if (Array.isArray(val0) && typeof(val0[0]) != 'undefined' && typeof(val1[0]) != 'undefined') {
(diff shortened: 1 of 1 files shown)


src/components/misc/personViews/PersonViewTable.jsx
```diff
@@ -128,19 +128,36 @@ export default class PersonViewTable extends React.Component {
 
                     // Sort, if a column is selected for sorting
                     if (this.state.sortIndex !== null) {
-                        visibleRows = visibleRows.concat().sort((row0, row1) => {
-                            const val0 = row0.data.content[this.state.sortIndex] || '';
-                            const val1 = row1.data.content[this.state.sortIndex] || '';
+                            visibleRows = visibleRows.concat().sort((row0, row1) => {
```
richardolsson, line 131: The indentation on this line (and hence all subsequent lines in this PR) is wrong. This line should be indented four spaces relative to the line before it, but is currently indented 8 spaces.

This means that all subsequent lines down to and including line 160 are indented four spaces too much.

richardolsson (changes requested):
The indentation is still wrong. 🙄 😊 

This is one of those things that will no longer be an issue when we have proper linting in place. But for now you need to fix it manually (or using some plugin in your editor).

fannysoderlund:
I see, is it correct now? Still cant choose to not add line 96 though

richardolsson (changes requested):
I'm afraid it's still not right. Line 96 is fine (it removes some unnecessary whitespace that was there before). And all of the lines that you've changed are correct amongst themselves, just not relative to the surrounding code.

See the difference between line 129 (existing code) and line 130 (your first change). Note how there is an 8-space difference between them. Add 8 spaces to 130-159 so that 130 and 129 are at the same level.

fannysoderlund:
Third time’s a charm!

src/components/misc/personViews/PersonViewTable.jsx
```diff
@@ -132,7 +132,24 @@ export default class PersonViewTable extends React.Component {
                             const val0 = row0.data.content[this.state.sortIndex] || '';
                             const val1 = row1.data.content[this.state.sortIndex] || '';
 
-                            let x = val0.toString().localeCompare(val1.toString());
+                            let x = 0;
+
+                            if (Array.isArray(val0) && typeof(val0[0]) != 'undefined' && typeof(val1[0]) != 'undefined') {
+                                x = val0[0].text.toLowerCase().toString().localeCompare(val1[0].text.toLowerCase().toString());
```
richardolsson, line 138: On this line you're doing `toLowerCase().toString()` before `localeCompare()`. I have a few comments on this.

## `toString()` is unnecessary
First, the `toLowerCase()` method exists on `String` and returns a `String`, so if that method is available you are already working with a string and there is no need to do `toString()` explicitly. This would leave you with the following instead:

```js
// Shorter without explicit tostring()
x = val0[0].text.toLowerCase().localeCompare(val1[0].text.toLowerCase());
```

Next it turns out that `toLowerCase()` is not necessary, and in fact [not desirable](https://stackoverflow.com/a/49004987/229824) in this scenario. Let's investigate why!

## Do we need `toLowerCase()`?
Without `localeCompare()` (which hasn't always existed in javascript), sorting an array of strings will yield a result where all uppercase strings come before the lowercase ones, like so:

```js
['abc', 'DEF', 'def'].sort();
// Result: ['DEF', 'abc', 'def']
````

It does this because the uppercase characters 'ABCDEFG…' come before the lowercase equivalents 'abcdefg…' in [the ASCII table](http://ascii-table.com/), which is the original lookup table used by computers to convert between binary data and human-readable characters (nowadays it's been extended by Unicode and other codecs to include more characters, symbols, emojis etc).

The default sorting behaviour is obviously undesirable, because any human would expect 'abc' to come before 'DEF'. A solution to this has been to use `toLowerCase()` so that all strings were compared as lower-case:

```js
['DEF', 'abc', 'def'].sort((s0, s1) => {
  if (s0.toLowerCase() > s1.toLowerCase()) return 1;
  if (s1.toLowerCase() > s0.toLowerCase()) return -1;
  else return 0;
})
// Result: ['abc', 'DEF', 'def']
```

This is better, but it still doesn't solve a very big problem, which is that we need to work with strings that are not restricted to the 26-character basic English alphabet, i.e. the characters listed in the ASCII table. Where do we sort a string that begins with 'å'? Towards the end, right? It's obvious for Swedes, but not for computers. By default javascript (and most environments) will treat it as an 'a', and not put it at the end of a sort as we wold expect.

Enter `localeCompare()`. It's main purpose is to handle localization, i.e. different languages. But it also has the added benefit that it will treat 'd' and 'D' as related characters. This means that sorting will be much more human-friendly:

```js
['def', 'abc', 'DEF'].sort((s0, s1) => s0.localeCompare(s1));
// Result: ['abc', 'def', 'DEF']
```

This is exactly what we'd like! So it turns out we don't need `toLowerCase()` at all.

## Is `localeCompare()` case insensitive?
Does this mean that `localeCompare()` ignores lowercase vs uppercase? No, it does not. It will always prioritize lower-case characters over upper-case ones. Sorting with `localeCompare()` will put a string beginning with 'd' before the same string beginning with 'D'. But it will never (in English or Swedish at least) put 'd' or 'D' before 'c' or 'C'.

## Ditching `toLowerCase()`
Knowing this (and provided that we don't care about true case-insensitive sorting, which I don't think we have to) we can rewrite this line of code to look like this:

```js
x = val0[0].text.localeCompare(val1[0].text);
```

This is of course shorter, but I think also easier to read.

src/components/misc/personViews/PersonViewTable.jsx
```diff
@@ -132,7 +132,24 @@ export default class PersonViewTable extends React.Component {
                             const val0 = row0.data.content[this.state.sortIndex] || '';
                             const val1 = row1.data.content[this.state.sortIndex] || '';
 
-                            let x = val0.toString().localeCompare(val1.toString());
+                            let x = 0;
+
+                            if (Array.isArray(val0) && typeof(val0[0]) != 'undefined' && typeof(val1[0]) != 'undefined') {
```
richardolsson, line 137: I think it would be easier to read if you separate the array check from the `undefined` check (also see comments on lines 144-152.

src/components/misc/personViews/PersonViewTable.jsx
```diff
@@ -132,7 +132,24 @@ export default class PersonViewTable extends React.Component {
...
+
+                            if(typeof(val0[0]) == 'undefined' && typeof(val1[0]) != 'undefined') {
+                                return 1;
+                            }
+                            if(typeof(val0[0]) != 'undefined' && typeof(val1[0]) == 'undefined') {
+                                return -1;
+                            }
+                            if(typeof(val0[0]) == 'undefined' && typeof(val1[0]) == 'undefined') {
+                                return 0;
+                            }
```
richardolsson, line 152: These should only happen if we are working with arrays. They could be enclosed in an if statement checking that `Array.isArray(val0) && Array.isArray(val1)`.

This is already what you're doing on line 137, so it would make sense to put all of your new code in the same if statement verifying that we are indeed working with arrays here.

richardolsson (changes requested):
Great work on this! Thanks for being so patient with indentation. :)

I've added some comments below on how I think the code can be made more concise. Would you like to change it? Let me know if you can't because of time or any other reason, and I will do it instead.

fannysoderlund:
Thanks, but you’re the one with patience. I can fix this today!  Från: Richard OlssonSkickat: den 25 januari 2021 08:40Till: zetkin/organize.zetk.inKopia: Fanny Söderlund; AuthorÄmne: Re: [zetkin/organize.zetk.in] Bug/entry level/sort survey response (#1231) @richardolsson requested changes on this pull request.Great work on this! Thanks for being so patient with indentation. :)I've added some comments below on how I think the code can be made more concise. Would you like to change it? Let me know if you can't because of time or any other reason, and I will do it instead.In src/components/misc/personViews/PersonViewTable.jsx:> @@ -132,7 +132,24 @@ export default class PersonViewTable extends React.Component {                              const val0 = row0.data.content[this.state.sortIndex] || '';                              const val1 = row1.data.content[this.state.sortIndex] || '';   -                            let x = val0.toString().localeCompare(val1.toString()); +                            let x = 0; + +                            if (Array.isArray(val0) && typeof(val0[0]) != 'undefined' && typeof(val1[0]) != 'undefined') { +                                x = val0[0].text.toLowerCase().toString().localeCompare(val1[0].text.toLowerCase().toString()); On this line you're doing toLowerCase().toString() before localeCompare(). I have a few comments on this.toString() is unnecessaryFirst, the toLowerCase() method exists on String and returns a String, so if that method is available you are already working with a string and there is no need to do toString() explicitly. This would leave you with the following instead:// Shorter without explicit tostring() x = val0[0].text.toLowerCase().localeCompare(val1[0].text.toLowerCase());Next it turns out that toLowerCase() is not necessary, and in fact not desirable in this scenario. Let's investigate why!Do we need toLowerCase()?Without localeCompare() (which hasn't always existed in javascript), sorting an array of strings will yield a result where all uppercase strings come before the lowercase ones, like so:['abc', 'DEF', 'def'].sort(); // Result: ['DEF', 'abc', 'def']It does this because the uppercase characters 'ABCDEFG…' come before the lowercase equivalents 'abcdefg…' in the ASCII table, which is the original lookup table used by computers to convert between binary data and human-readable characters (nowadays it's been extended by Unicode and other codecs to include more characters, symbols, emojis etc).The default sorting behaviour is obviously undesirable, because any human would expect 'abc' to come before 'DEF'. A solution to this has been to use toLowerCase() so that all strings were compared as lower-case:['DEF', 'abc', 'def'].sort((s0, s1) => {   if (s0.toLowerCase() > s1.toLowerCase()) return 1;   if (s1.toLowerCase() > s0.toLowerCase()) return -1;   else return 0; }) // Result: ['abc', 'DEF', 'def']This is better, but it still doesn't solve a very big problem, which is that we need to work with strings that are not restricted to the 26-character basic English alphabet, i.e. the characters listed in the ASCII table. Where do we sort a string that begins with 'å'? Towards the end, right? It's obvious for Swedes, but not for computers. By default javascript (and most environments) will treat it as an 'a', and not put it at the end of a sort as we wold expect.Enter localeCompare(). It's main purpose is to handle localization, i.e. different languages. But it also has the added benefit that it will treat 'd' and 'D' as related characters. This means that sorting will be much more human-friendly:['def', 'abc', 'DEF'].sort((s0, s1) => s0.localeCompare(s1)); // Result: ['abc', 'def', 'DEF']This is exactly what we'd like! So it turns out we don't need toLowerCase() at all.Is localeCompare() case insensitive?Does this mean that localeCompare() ignores lowercase vs uppercase? No, it does not. It will always prioritize lower-case characters over upper-case ones. Sorting with localeCompare() will put a string beginning with 'd' before the same string beginning with 'D'. But it will never (in English or Swedish at least) put 'd' or 'D' before 'c' or 'C'.Ditching toLowerCase()Knowing this (and provided that we don't care about true case-insensitive sorting, which I don't think we have to) we can rewrite this line of code to look like this:x = val0[0].text.localeCompare(val1[0].text);This is of course shorter, but I think also easier to read.In src/components/misc/personViews/PersonViewTable.jsx:> @@ -132,7 +132,24 @@ export default class PersonViewTable extends React.Component {                              const val0 = row0.data.content[this.state.sortIndex] || '';                              const val1 = row1.data.content[this.state.sortIndex] || '';   -                            let x = val0.toString().localeCompare(val1.toString()); +                            let x = 0; + +                            if (Array.isArray(val0) && typeof(val0[0]) != 'undefined' && typeof(val1[0]) != 'undefined') { I think it would be easier to read if you separate the array check from the undefined check (also see comments on lines 144-152.In src/components/misc/personViews/PersonViewTable.jsx:> +                            if(typeof(val0[0]) == 'undefined' && typeof(val1[0]) != 'undefined') { +                                return 1; +                            } +                            if(typeof(val0[0]) != 'undefined' && typeof(val1[0]) == 'undefined') { +                                return -1; +                            } +                            if(typeof(val0[0]) == 'undefined' && typeof(val1[0]) == 'undefined') { +                                return 0; +                            } These should only happen if we are working with arrays. They could be enclosed in an if statement checking that Array.isArray(val0) && Array.isArray(val1).This is already what you're doing on line 137, so it would make sense to put all of your new code in the same if statement verifying that we are indeed working with arrays here.—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.

richardolsson (approved):
Great work! The code looks much nicer now, and it still works as expected. Thanks for participating in this iterative process with me. 😊 🙌

PR #1232, opened 2021-01-24
Collapse view header

### LINKED ISSUE #1225: Collapse view "header"
Most of the work in a view does not require constantly changing the name/description or switching between tabs. We should add an option to collapse (hide/minimize) these similarly to how Google Documents allow the user to minimize the toolbar.

niklasva82:
Resolves #1225

### DIFF
--- a/src/components/sections/people/PersonViewsPane.jsx
+++ b/src/components/sections/people/PersonViewsPane.jsx
@@ -36,6 +36,7 @@ export default class PersonViewsPane extends RootPaneBase {
 
         this.state = {
             viewMode: 'saved',
+            collapseHeader: false,
         };
     }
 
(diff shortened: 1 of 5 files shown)


niklasva82:
collapse

richardolsson:
Conceptually this works great IMO, but I wonder if we should put the "collapse" link at the the rightmost edge of the screen instead?

Also it isn't properly handling the Smart Search "tab" (see screenshot below). I think it would make sense to show the header again if the user switches from the "Saved" tab to the "Smart Search" tab.

If the user hides the header while already in the Smart Search tab, maybe just hide the `RelSelectInput`? I'm not sure what's best. What do you think?

image

richardolsson (changes requested):
Issue described above needs to be fixed.

olleolleolle:
I want to share the happy sentiment: THIS CONCEPT IS A SUPER 👋  IMPROVEMENT 👋

niklasva82:
All the issues above should be addressed now.

src/components/sections/people/PersonViewsPane.jsx
```diff
@@ -151,7 +164,9 @@ export default class PersonViewsPane extends RootPaneBase {
                             value={ this.state.query } objects={ queries } showEditLink={ true }
                             onValueChange={ (name, val) => this.setState({ query: val }) }
                             onCreate={ this.onQueryCreate.bind(this) }
-                            onEdit={ this.onQueryEdit.bind(this) }/>
+                            onEdit={ this.onQueryEdit.bind(this) }
+                            hidden={ this.state.collapseHeader }
```
richardolsson, line 168: I would have preferred removing the `RelSelectInput` entirely instead of just hiding it with CSS. But that's a matter of taste, and soon enough we don't have to worry about things like that and instead let the UI library decide for us. For now, lets keep it like this. 😊

richardolsson (approved):
Code looks good and is working as epxected. This is a really nice improvement!

PR #1233, opened 2021-01-28
PersonNotes column in PersonView

niklasva82:


src/components/misc/personViews/cells/PersonNotesCell.jsx
```diff
@@ -0,0 +1,22 @@
...
+
+export default function PersonNotesCell(props) {
+    let notes;
+    if (props.content && props.content.length) {
+        const lines = props.content.filter(line => !!line.text && !!line.text.length);
+        notes = lines.map(n => (
+            <div key={ n.id }
+                className="PersonNotesCell-note"
+                onClick={ () => props.openPane('notes', 'person', n.person_id) } >
+                { n.text.replace(/<[^>]*>/g, '') }
```
richardolsson, line 12: Nice solution to handling styled notes!

src/components/misc/personViews/cells/PersonNotesCell.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import truncText from '../../../../utils/truncText';
```
richardolsson, line 2: I assume you imported this with the goal of truncating the text for long notes, the same way that `SurveyResponseCell` component truncates long responses. I think `PersonNotesCell` should too, but right now this import is not actually being used in this component, and long notes are not being truncated.

richardolsson (changes requested):
Great work! The code looks great and things are working mostly as expected. I'd like long notes to be truncated, I think you intended to implement that but somehow lost track of it along the way? 😊 

Please implement truncation and then I'm ready to merge!

src/components/misc/personViews/cells/PersonNotesCell.jsx
niklasva82: Fixed

richardolsson (approved):
Nice! Looks good to me and works as expected! 👍 

I made a small improvement myself which I felt it was easier to just do than to try and explain. Views containing this column will now update when a person in the view receives a new note.

See a9cd09312f9a491ea53cf5e3b5120f730e1d24fc for details.

PR #1234, opened 2021-01-28
View column reordering

### LINKED ISSUE #1113: View column re-ordering
Add drag-and-drop interface to re-order view columns.

richardolsson:
This PR adds an interface to reorder view columns. I found it really difficult to achieve this reliable _within_ the view, so I added a new pane to do it. This will also be the place for future settings such as access/sharing, ownership and other settings. We could potentially move "delete" here as well.

Finally, it changes the way view column sorting works so that you can click anywhere on the column head to sort it. This is made possible by moving the column configuration to the new settings pane. Clicking the column head used to open column settings, but now you need to go through the new pane to edit the column.

* Add new View settings pane
* Implement view column reordering (requires changes to Platform API)
* Click column in settings pane to edit column settings
* Click anywhere on column header in view to sort

image

Fixes #1113


niklasva82 (approved):
Looks good to me!

PR #1235, opened 2021-01-28
Tweak collapse/back links in person view

richardolsson:
This PR tweaks the "back" and "collapse" links for person views:

* Makes the entire link clickable, including the chevron icon
* Moves the chevron of the "collapse" link to the right-hand side
* Animates chevron by rotating it (see animation below)

view-collapse-rotate


niklasva82 (approved):
Looks good!

PR #1237, opened 2021-01-29
Survey submitted view column

richardolsson:
This PR adds a new view column type for survey submission date. It will render the date of the most recent submission to a given survey.

image


niklasva82 (approved):
Looks good and works for me!

PR #1238, opened 2021-02-02
Local view columns (person and bool)

### LINKED ISSUE #1236: View column for person (e.g. "assignee")
Implement the soon to be introduced `local_person` view column type. This can be used, for instance, to assign some task related to a member to a certain official. Two use cases:

* A view containing people to be mobilized for some task, e.g. a call assignment. The `local_person` column type could be add to assign each member to a particular organizer in charge of reaching out to them and reporting back.
* A view containing union members with open cases. The `local_person` column can be used to indicate which case worker is assigned to each of the cases.

Clicking a cell should open up a dialog where the user can find the person they're trying to assign.

richardolsson:
This PR implements the recently introduced `local_bool` and `local_person` view column types.

* Create UI for adding and displaying the `local_bool` column type
* Create UI for adding and displaying the `local_person` column type
  * Create pop-over to select person or open pane for current person
* Switch the icon used in `person_tag` boolean to distinguish from `local_bool`
* Improve view sorting algorithm

image

Closes #1236


niklasva82:
First glance it looks great! But I would like for the input field to be automatically focused when I click the person field so I can start typing immediately :)

niklasva82 (changes requested):
Looks great! And I love the new tag icon as well to differentiate between tags and local values. I would like a change as described above, for the input field to be focused automatically when opening the select person dialog.

richardolsson:
> First glance it looks great! But I would like for the input field to be automatically focused when I click the person field so I can start typing immediately :)

I've added auto-focus to that input now. Let me know what you think!

There are a lot of other keyboard navigation improvements that could be made, but that we need to save for later 😊

niklasva82 (approved):
Great! Another change that would be awesome is keyboard navigation to select results. But I will open a separate issue for that :)

PR #1242, opened 2021-02-07
WIP: Add scrollListToSelectedIndex to RelSelectInput

olleolleolle:
This PR is the exploration @oelrich and I did on the RelSelectInput component, which needs a "memory" of the last selected option, to be "in viewport" when the options are re-shown.

## Background

This new state _is meant_ to make it possible to modify scrollTop of the listDOMNode - the UL element - when a selection has been made, previously, and the user is re-opening the "selectable options" list. It doesn't do that!

## Problems, questions

- When can the scrollTop be set?


richardolsson:
Great work on this! @niklasva82 ended up implementing the same thing using a slightly different approach in #1275.

PR #1249, opened 2021-02-23
Bugfix/export arrays

### LINKED ISSUE #1243: View column survey response not serialized properly in export
The view column survey response is serialized as [object Object] on export from view.

niklasva82:
Serialize arrays and person objects on export from view.

Resolves #1243

### DIFF
--- a/src/actions/personView.js
+++ b/src/actions/personView.js
@@ -105,8 +105,14 @@ export function exportPersonView(viewId, queryId) {
 
             // Add all rows
             rowList.items.forEach(rowItem => {
-                const data = rowItem.data;
-                rows.push([data.id].concat(data.content));
+                let content = rowItem.data.content;
+                if(typeof(content == 'object') {
+                    if(content instanceof Array) {
+                        content = content.join("\n");
+                    }
+                }
+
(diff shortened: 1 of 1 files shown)


src/actions/personView.js
```diff
@@ -106,7 +106,30 @@ export function exportPersonView(viewId, queryId) {
             // Add all rows
             rowList.items.forEach(rowItem => {
                 const data = rowItem.data;
-                rows.push([data.id].concat(data.content));
+                let content = data.content;
+
+                content = content.map(item => {
```
richardolsson, line 111: Couldn't this just be `const content = data.content.map(item => {`?

richardolsson (changes requested):
I found (I think) a minor improvement to the code. But I also wanted to check with you: Could we take this opportunity to add a case to handle booleans as well? It would be nice to have them come out as `'x'` (for `true`) or `''` (i.e. empty string for `false`).

What it currently looks like:
image

What would be nice (more readable IMO):
image

richardolsson (approved):
Thanks! This looks very good now. Great improvement! 🙌

PR #1252, opened 2021-03-03
Disable autocomplete on forms

### LINKED ISSUE #1248: Disable form autocomplete
One user reports that her browser autofill feature caused all the information in a form to be overwritten. We should do what we can to disable browser auto-complete.

MDN contains [more info on how to do this](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion), but there may be other variations or solutions that apply.

niklasva82:
Resolves #1248

### DIFF
--- a/src/components/forms/Form.jsx
+++ b/src/components/forms/Form.jsx
@@ -28,7 +28,8 @@ export default class Form extends React.Component {
 
         return (
             <form className={ classes }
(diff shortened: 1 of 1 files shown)


richardolsson (approved):
LGTM! 💯

PR #1254, opened 2021-03-09
Add gender view column to available options and a gender view cell

### LINKED ISSUE #1245: "Gender" not available as field when creating view column
Steps to reproduce:

1. Navigate to People > Views
2. Create a view
3. Click "Add Column"
4. Select "Person Field"

Expected result: The "Gender" field should be available in the select list
Actual result: "Gender" is missing from the list

image

niklasva82:
Resolves #1245


richardolsson (changes requested):
Almost there! Everything looks good, execpt for two missing strings. 🤏

richardolsson (approved):
Great! This is working now. :)

PR #1255, opened 2021-03-10
Render gender properly in JoinSubmissionPane

### LINKED ISSUE #1247: Gender not rendered correctly by JoinSubmissionPane
Looks like the `gender` field is not being treated as a native field, but instead as a custom field (which doesn't exist, so the label is empty). See screenshot.

image

niklasva82:
Resolves #1247

### DIFF
--- a/locale/misc/fields/en.yaml
+++ b/locale/misc/fields/en.yaml
@@ -11,8 +11,7 @@ phone: Phone number
 street_address: Street address
 zip: Zip code
 genderOptions:
-  f: Female
-  m: Male
-  o: Other
-  null: Unknown
-  _: Unknown
+  female: Female
+  male: Male
+  other: Other
+  unknown: Unknown
--- a/locale/misc/fields/en.yaml
+++ b/locale/misc/fields/en.yaml
@@ -10,3 +10,7 @@ last_name: Last name
 phone: Phone number
 street_address: Street address
 zip: Zip code
+genderOptions:
+  f: Female
+  m: Male
+  o: Other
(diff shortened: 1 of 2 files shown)


locale/misc/fields/en.yaml
```diff
@@ -10,3 +10,9 @@ last_name: Last name
 phone: Phone number
 street_address: Street address
 zip: Zip code
+genderOptions:
+  f: Female
+  m: Male
+  o: Other
+  null: Unknown
+  _: Unknown
```
richardolsson, line 18: There should never be a case where the value (from the API) is `_`. The underscore is only used in some UI implementations because it's not possible to store `null` as a value in HTML attributes.

If there is a case where `_` occurs in the data from the API, that's a bug. Have you seen that happen?

src/components/panes/JoinSubmissionPane.jsx
```diff
@@ -151,7 +152,15 @@ export default class JoinSubmissionPane extends PaneBase {
                             label = this.props.intl.formatMessage({ id: `misc.fields.${fieldName}` });
                         }
 
-                        const value = person[fieldName] || '';
+                        let value;
+                        if(fieldName == 'gender') {
+                            value = this.props.intl.formatMessage({ id: `misc.fields.genderOptions.${person[fieldName]}` })
```
richardolsson, line 157: If we are worried that the value is not one of the four permitted values (`'f'`, `'m'`, `'o'`, `null`) we should probably fall back anything that isn't f/m/o to some other value and use that as the message ID (e.g. `'misc.fields.genderOptions.unknown'`).

richardolsson (changes requested):
I'm curious why you've added a fifth gender option here. Have you found a bug, or is this a mistake?

richardolsson (approved):

PR #1258, opened 2021-03-11
Fix custom field query rendering bug

### LINKED ISSUE #1250: Bug rendering custom field in query
Steps to reproduce:

1. Create a Smart Search query or all assignment
2. Open/edit the query
3. Add a filter for "Custom Field"
4. Select a custom field (I experienced this with a date field, but should be same for any type)
5. Configure the filter
6. Save and close the Smart Search
7. Refresh the page (to clear state)
8. Open the edit pane for the query again

Expected results: Should look exactly like after step 5
Actual results: The "Choose field" select box is the only one showing, probably because the list of fields had not loaded when the filter list was rendered

niklasva82:
Fix bug where PersonFieldFilter will not render properly if fieldTypes is not available when class constructor is run.

Resolves #1250

### DIFF
--- a/src/components/filters/PersonFieldFilter.jsx
+++ b/src/components/filters/PersonFieldFilter.jsx
@@ -37,6 +37,15 @@ export default class PersonFieldFilter extends FilterBase {
         }
     }
 
+    componentWillReceiveProps(nextProps, nextState) {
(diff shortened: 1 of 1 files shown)


richardolsson (approved):
Nice! Clean and simple fix. 🙌

PR #1259, opened 2021-03-11
Update view when relevant person object updated

### LINKED ISSUE #1256: Update view when person updates
Reload view row when a person field updates, provided that the field is used in the view.

Similar logic exists in the reducer for `personView` for a number of other changes, e.g. when a tag is toggled.

niklasva82:
Resolves #1256

### DIFF
--- a/src/store/personViews.js
+++ b/src/store/personViews.js
@@ -419,13 +419,6 @@ export default function personViews(state = null, action) {
             dirtyPersonIds.push(pId);
 
             Object.keys(state.rowsByView)
-                .filter(viewId => {
-                    // Filter views that contain the dirty person object
-                    const rowList = state.rowsByView[viewId];
-                    if (rowList.items) {
-                        return rowList.items.find(row => row.data.id === pId)
-                    }
-                })
                 .forEach(viewId => {
                     // Add any affected views to list
                     affectedViewIds.push(viewId)
--- a/src/store/personViews.js
+++ b/src/store/personViews.js
@@ -414,6 +414,23 @@ export default function personViews(state = null, action) {
             dirtyPersonIds.push(action.meta.rowId);
             affectedViewIds.push(action.meta.viewId);
         }
+        else if (action.type == types.UPDATE_PERSON + '_FULFILLED') {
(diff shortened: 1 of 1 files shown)


richardolsson (approved):
Looking great and works as expected!

src/store/personViews.js
```diff
@@ -414,6 +414,23 @@ export default function personViews(state = null, action) {
...
+            dirtyPersonIds.push(pId);
+
+            Object.keys(state.rowsByView)
+                .filter(viewId => {
+                    // Filter views that contain the dirty person object
+                    const rowList = state.rowsByView[viewId];
+                    if (rowList.items) {
+                        return rowList.items.find(row => row.data.id === pId)
+                    }
+                })
```
richardolsson, line 428: The corresponding logic for other actions don't just check for views containing a person, but views that may be affected by the update. In the case of a taging for example, on lines 355-377, it looks for views that contain a tag column.

Because what we're working with here is a person object changing, any view may have updated in response to this. So I think there should be no filtering happening here at all. The next pass starting at line 437 will go through all of the views and flag them if the person exists either in a saved row, or in a row retrieved dynamically using a query.

richardolsson (changes requested):
Actually I have to take that back! This is not working for for Smart Search query view tabs. Steps to reproduce:

1. Create a view
2. Load a Smart Search query into the view
3. With the Smart Search tab of the view selected, open any person from the view *that is not starred/saved*
4. Edit the person and save

Expected results: The row should update
Actual results: Nothing happens to the row

src/store/personViews.js
richardolsson: Just to clarify: When I say that "any view" will have been affected, that's not strictly true, because not all views necessarily contain a `person_field` column. But in practice, they always do, because Zetkin Organize adds first name and last name by default, and it's extremely rare that someone would delete them and not add any other `person_field` columns.

niklasva82:
Seems to work now that I removed filtering

richardolsson (approved):
Confirmed to be working both for saved rows and Smart Search queries. ✅

PR #1261, opened 2021-03-16
Delete survey submission

### LINKED ISSUE #665: Delete survey submission
It should be possible to delete a survey submission, e.g. when there are duplicates or when someone submits an empty survey.

niklasva82:
* Adds action and reducer for deletion of survey submission.
* Adds delete link to survey submission which opens ConfirmDeletePane.
* Extends ConfirmDeletePane to handle survey submissions.

Resolves #665


richardolsson (changes requested):
It works very well! The only thing that isn't great is that when you delete a survey submission, the pane containing that submission does not close. This is handled for many other scenarios, and is best implemented in the `view` reducer to find and close any panes representing the newly deleted submission, by ID.

Could you please implement that?

richardolsson:
In case anyone's interested, here is how I tested that this works and does what one would expected.

I set up a call assignment to quickly respond to the survey that is part of the dummy data, on behalf of six dummy people. I then created three Smart Search queries, one of each type that is related to survey submissions: `survey_submission`, `survey_response` and `survey_option`.

I then set up a view containing columns for the three smart searches, and one of each of the view columns that relate to survey submissions. I added all the six respondents to the view.

I then tried to delete two submissions, and looked at the view to verify that all indications of the submissions ever having existed were no longer in the view:

image

This way, it tests both the deleting itself, but also the survey-related Smart Search and view features.

richardolsson (approved):
LGTM!

PR #1262, opened 2021-03-20
Fixes view scroll bug

### LINKED ISSUE #1253: Cannot scroll right on view header without people in view
To reproduce:
1. Create new view
2. Add as many columns as needed for scrolling to be required
3. Try srolling right
Expected: Scroll works
Actual: Scroll does not work

The scroll works once people are added to the view however.

deSelbyPsyche:
Adds overflow css so that the window can be scrolled horizontally on a large screen as well as a small screen, fixes #1253

### DIFF
--- a/src/components/misc/personViews/PersonViewTable.scss
+++ b/src/components/misc/personViews/PersonViewTable.scss
@@ -105,6 +105,7 @@
         right: 0;
         left: 0;
         bottom: 0;
+        overflow: auto;
 
         tbody {
             position: absolute;


niklasva82 (changes requested):
Great job! But there is a bug here.

The fix works as expected when no people have been added, but as soon as people are added to the view and scroll is needed, two scroll bars appear.

This appears to be because the existing scroll is on the table body, and the new scroll is on table. If you remove the scroll from the table body, the problem will likely be resolved.

richardolsson:
> This appears to be because the existing scroll is on the table body, and the new scroll is on table. If you remove the scroll from the table body, the problem will likely be resolved.

The reason why the scrollbar was on `tbody` (and the root cause of this bug) is that we want to be able to scroll the body of the table vertically without the header scrolling out of view, i.e. a "sticky header". As far as I know, that can't be achieved without a JS hack like the one employed here, where the header is scrolled programmatically as an effect of the user scrolling the body.

So I think the strategy applied here might not be the right way to achieve a solution. But it would be interesting to see if having vertical scrolling on body and horizontal scrolling on the entire table would be an acceptable solution. I'm not sure how the browsers would handle that, but it's worth a try! 😊

richardolsson:
Closing this as stale. This issue does not exist in Gen3.

PR #1263, opened 2021-03-20
Fix AddViewColumn bug

niklasva82:
Fix undocumented bug where adding a view column without selecting anything would add a person_field filter with the configuration of a person_note.

To reproduce:
1. Create a view
2. Add a few people to the view
3. Click Add column
4. Click the Add column button in the AddViewColumnPane

Expected results: Adds the selected view column type
Actual results: Adds a column named Person notes, clears people from the view due to the backend not knowing what to do with the odd column configuration.

### DIFF
--- a/src/components/panes/AddViewColumnPane.jsx
+++ b/src/components/panes/AddViewColumnPane.jsx
@@ -238,16 +238,18 @@ class PersonFieldColumnTemplate extends React.Component {
 class PersonNotesColumnTemplate extends React.Component {
 
     componentDidMount() {
-        const column = {
-            title: this.props.intl.formatMessage({
-                id: 'panes.addViewColumn.templates.person_notes.title'
-            }),
-            config: {
-                limit: 1,
-            }
-        };
+        if(this.props.selected) {
+            const column = {
+                title: this.props.intl.formatMessage({
+                    id: 'panes.addViewColumn.templates.person_notes.title'
+                }),
+                config: {
+                    limit: 1,
+                }
+            };
 
-        this.props.onChange(column);
+            this.props.onChange(column);
+        }
     }
 
     render() {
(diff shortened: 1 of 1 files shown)


src/components/panes/AddViewColumnPane.jsx
```diff
@@ -238,16 +238,18 @@ class PersonFieldColumnTemplate extends React.Component {
...
     componentDidMount() {
-        const column = {
-            title: this.props.intl.formatMessage({
-                id: 'panes.addViewColumn.templates.person_notes.title'
-            }),
-            config: {
-                limit: 1,
-            }
-        };
+        if(this.props.selected) {
```
richardolsson, line 241: This will never be true, because the component only mount once, and it's not `selected` when it first mounts. It *might* work putting this in a `componentWillRecieveProps()` instead.

richardolsson (changes requested):
Great that you discovered this and took care of it straight away! We actually had a user report this yesterday, and I immediately recognized it. 😊 

Unfortunately, this fix does not seem to work as it introduces a new bug. Steps to reproduce:

1. Create a new view
2. Click "Add column"
3. Select the "Person notes" column type
4. Click the blue _Add column_ button

Expected results: A new column of type `person_notes` should be added and named _Person notes_ or something like that
Actual results: A new column of the right type is added, but it's named _E-mail_ (or whatever was the first selected field in the `person_field` column config. See screenshot:

image

I've added a comment below on how I think this could be fixed, but please confirm:

niklasva82:
In addition to fixing the above bug, this now also fixes a bug where switching between survey response and survey submission would result in weird behavior.

richardolsson (approved):
Awesome! Works as expected and the code looks good. 🎉

PR #1264, opened 2021-03-20
Allows fuzzy loading of names

### LINKED ISSUE #1246: PersonSelectWidget not supporting fuzzy search
`PersonSelectWidget` does not support fuzzy search, even though the search API used does. Steps to reproduce:

1. Navigate to People > Views
2. Create a view
3. At the bottom of the view, type into the search field a misspelled name, e.g. "Chole" (for Chloe)

Expected results: Should find and suggest dummy persons named Chloe
Actual results: Finds nothing

This is likely because `PersonSelectWidget` uses `RelSelectInput`, which performs client-side filtering for more rapid updates on type, and that filtering does not support fuzzy search.

I propose that the simplest solution is to allow `RelSelectInput` filtering to be disabled, and to do so for `PersonSelectWidget`, instead relying entirely on server-side search.

deSelbyPsyche:
…-side fuzzy loading, fixes #1246

### DIFF
--- a/src/components/forms/inputs/RelSelectInput.jsx
+++ b/src/components/forms/inputs/RelSelectInput.jsx
@@ -39,12 +39,13 @@ export default class RelSelectInput extends InputBase {
 
     renderInput() {
         const value = this.props.value;
-        const objects = this.props.objects;
+        let objects = this.props.objects;
         const showEditLink = this.props.showEditLink;
         const valueField = this.props.valueField;
         const selected = (value && objects)?
             objects.find(o => o[valueField] == value) : null;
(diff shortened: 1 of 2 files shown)


richardolsson (approved):
Flawless

Nice work! No objections here, and I've verified that it does what it says on the tin. First pull request approved! 🎉 

Only minor comment that I have is on the naming of the branch: It's missing the dash `-` (not important) and there is a typo in the issue number (slightly more important), but that's not something we need to care about as soon as I've merged it, which I'm going to do now! 👏 🙌

PR #1265, opened 2021-04-07
Disconnect person instead of deleting

niklasva82:
Implement interface for disconnecting a person instead of deleting.

### DIFF
--- a/locale/panes/confirmDelete/en.yaml
+++ b/locale/panes/confirmDelete/en.yaml
@@ -1,4 +1,6 @@
 confirmButton: Delete
+deleteConnection: Do you want to remove this person from your organization?
+deleteConnectionOrgs: 'This will also remove the person from these organizations:'
 deleteView: Do you want to delete this view?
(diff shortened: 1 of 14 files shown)


richardolsson (changes requested):
I've tried this and it looks great to me! When I was about to take a closer look at the code I noticed a problem which is that you've requested to merge it into master, but it should be merged to (and branched from) `orgsharing`.

I tried just changing the base to `orgsharing` but the commit diff didn't look right. Did you perhaps accidentally branch this off `master` when you created the branch?

Finally, and this is a tiny thing. The branch name references an issue which is not part of this repository. In the context of this repo, I believe this issue is undocumented.

richardolsson (approved):
Great! Works as expected and code looks great (except only for some minor indentation issues, but let's ignore that for this legacy codebase 😊).

niklasva82:
Ah, probably because I installed a new distro over the weekend and it took
a day or so before I migrated my vimrc 😊

Den ons 7 apr. 2021 17:18Richard Olsson ***@***.***> skrev:

PR #1268, opened 2021-04-20
URL in Action

niklasva82:
Add action field to ActionForm and ActionPane.
Normalize URL on blur. I had to use an old version of the normalize-url package for this due to an incompatibility with a feature in newer ES version than the old version of babel can handle (...)

### DIFF
--- a/src/components/forms/ActionForm.jsx
+++ b/src/components/forms/ActionForm.jsx
@@ -1,5 +1,7 @@
 import React from 'react';
 
+import normalizeUrl from 'normalize-url';
+
 import Form from './Form';
 import IntInput from './inputs/IntInput';
 import DateInput from './inputs/DateInput';
@@ -71,9 +73,16 @@ export default class ActionForm extends React.Component {
                     initialValue={ action.title } maxLength={ 300 }/>
                 <TextArea labelMsg="forms.action.info" name="info_text"
                     initialValue={ action.info_text }/>
-
+                <TextInput labelMsg="forms.action.url" name="url"
+                    initialValue={ action.url } maxLength={ 500 }
+                    onBlur={ this.onChangeUrl.bind(this) }/>
             </Form>
         );
+    }
+
+    onChangeUrl(ev) {
+        const url = ev.target.value;
+        ev.target.value = normalizeUrl(url);
     }
 
     getValues() {
--- a/src/components/forms/ActionForm.jsx
+++ b/src/components/forms/ActionForm.jsx
@@ -1,13 +1,12 @@
 import React from 'react';
 
-import normalizeUrl from 'normalize-url';
-
 import Form from './Form';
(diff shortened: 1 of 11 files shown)


src/components/forms/ActionForm.jsx
```diff
@@ -71,11 +73,18 @@ export default class ActionForm extends React.Component {
...
+                <TextInput labelMsg="forms.action.url" name="url"
+                    initialValue={ action.url } maxLength={ 500 }
+                    onBlur={ this.onChangeUrl.bind(this) }/>
             </Form>
         );
     }
 
+    onChangeUrl(ev) {
+        const url = ev.target.value;
+        ev.target.value = normalizeUrl(url);
```
richardolsson, line 85: This does not feel very "reacty". You should never change an input field like this, especially if it's a "controlled" input, i.e. an input that get it's value from the declarative JSX, as is the case in TextInput.jsx:28.

src/components/forms/inputs/InputBase.jsx
```diff
@@ -37,9 +37,14 @@ export default class InputBase extends React.Component {
...
-    }
+    } 
 
     onChange(ev) {
         this.props.onValueChange(this.props.name, ev.target.value);
     }
+
+    onBlur(ev) {
+        this.props.onBlur(ev);
+        this.props.onValueChange(this.props.name, ev.target.value);
```
richardolsson, line 48: I'm guessing the only reason this is working (if it actually is working?) is because you are calling `onValueChange()` after blur, which feels weird to me. Nothing should have changed on blur (except maybe on the next `render()`).

richardolsson (changes requested):
It looks to me like this code circumvents React and makes changes directly to the DOM, which is bad practice. Did you try it any other way?

I would have assumed that the best way to do it would be to normalize the URL right before the patch/post to the API (you can even do it before calling back from the form, or create a new input component for URLs and handle it there).

richardolsson (changes requested):
This seems to work now, and looks good! But can you please rearrange the branches so that we can merge this onto `orgsharing` instead?

niklasva82:
Ah sorry I thought I did that yesterday!

richardolsson:
Please have a look at the PR diff now, and you will see that it now includes commits that you probably do not want to treat as part of this PR, @niklasva82. I wasn't just asking to change the base of the PR (I can do that myself) but also make sure that the branches are in sync. Is there nothing else that needs to be done before we merge this?

niklasva82:
Yeah sorry, here is the PR for merging matser to orgsharing which is a good idea to do before this. #1270

richardolsson (approved):
Thanks for making the necessary preparations!

Note how the PR commit log only updates when the base is changed, so it's good practice to first make sure the base is up to date (i.e. merge `master` into `orgsharing` in this case) and only then update the base on this one.

PR #1269, opened 2021-04-23
Orgpicker scrollbar

niklasva82:
Add scrollbar to organization picker.


richardolsson (approved):
This seems to work and looks ok for now! For Generation 3 we will create something nicer.

PR #1272, opened 2021-04-30
Include slug in export file name

### LINKED ISSUE #1260: Include org slug in view export file name
When exporting a view, the file is named according to the date and time of the export, e.g. `20210311_120307.csv`.

For someone who works with multiple organizations (or even sub-organizations) it would be useful to include the organization slug in the name, e.g.: `my-organization_20210311_120307.csv`.

niklasva82:
Resolves #1260 

Does not include slug exactly because it's not available without making changes to the way the platform returns organizations, instead merely returns a lowercased organization title stripped of any non-a-z-characters.


richardolsson (approved):
LGTM!

PR #1274, opened 2021-05-06
Fix reset of timeframe dates

### LINKED ISSUE #1213: Date reset after changing timeframe in filter
Steps to reproduce:

1. Create a Smart Search Query
2. Add a Campaign participation filter
3. Change the timeframe to "After a certain date"
4. Change the date (below) to 1857-07-05
5. Change the timeframe to "Between certain dates"

Expected results: Another date field should appear, but the 1857-07-05 date should remain in the existing one
Actual results: Another date field appears, and the existing one is reset to today's date

This affects all filters which use this timeframe configuration component.

image

niklasva82:
Resolves #1213

### DIFF
--- a/src/components/filters/CallHistoryFilter.jsx
+++ b/src/components/filters/CallHistoryFilter.jsx
@@ -196,6 +196,10 @@ export default class CallHistoryFilter extends FilterBase {
             this.onConfigChange());
     }
 
+    validDate(dateString) {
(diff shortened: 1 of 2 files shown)


richardolsson (approved):
LGTM! I've looked at the code and confirmed through testing that the bug was fixed.

PR #1275, opened 2021-05-06
Scroll to selected element in RelSelectInput after focus

### LINKED ISSUE #1240: Scroll RelSelectInput to currently selected item
The problem is illustrated below
scroll

To reproduce:
1. Create a few queries to fill out the list.
2. Select a query that is outside the view of the scroll window.
3. Edit the current query

Expected behavior:
Automatically scroll to position of currently selected item so that user can click the edit button.

Actual behavior:
The user has to find the currently selected query by scrolling, which can be quite painstaking when there are many queries, or press backspace and search again.

niklasva82:
Automatically scroll to the currently selected element in the RelSelectInput when focused.

Resolves #1240

### DIFF
--- a/src/components/forms/inputs/RelSelectInput.jsx
+++ b/src/components/forms/inputs/RelSelectInput.jsx
@@ -27,6 +27,13 @@ export default class RelSelectInput extends InputBase {
         }
     }
 
+    componentDidUpdate(prevProps, prevState) {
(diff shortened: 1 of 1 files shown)


richardolsson (approved):
Beautiful solution! Code looks god and I've confirmed that it works as expected. ✅ ✅ ✅ 🎉

PR #1281, opened 2021-06-10
Shared queries

niklasva82:
Changes in UI to reflect read-only queries shared from another organization.

The query reducer computes isShared for each query, determining if the query is shared from another organization.

The RelSelectInput shows a share icon instead of edit link when an item isShared.
relselect-shared

The PersonQueryListItem shows share icon and organization name if query isShared.
Skärmbild från 2021-06-10 02-10-35

The PersonQueryPane shows a label without edit link if query isShared.
Skärmbild från 2021-06-10 02-12-40

### DIFF
--- a/locale/panes/query/en.yaml
+++ b/locale/panes/query/en.yaml
@@ -1,4 +1,5 @@
 editLink: Edit this Smart Search query
+shared: 'Query is shared from {org}'
 summary:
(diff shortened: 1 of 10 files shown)


richardolsson:
No longer relevant. This has been implemented in other PRs.

Issue #1282, opened 2021-06-13
Server-side data-fetching fetches data from wrong organization when on non-default organization.


niklasva82:
The server-side data-fetching does not appear to work properly when in an organization other than the default (first) organization. The data fetched is not from the active organization, but instead is from the first organization in the list.

To reproduce:
1. Select an organization other than the first one.
2. Go to People. Note that all people have an avatar and are part of the organization. These have been loaded by the client.
3. Reload

Expected result: The list should look the same as in 2.
Actual results: The list includes people from the first organization.

The same can be repeated for other sections like actions, where actions and campaigns from the default organization are loaded server-side.

richardolsson:
Duplicate of #798

PR #1283, opened 2021-06-13
Workaround for #1282

niklasva82:
Turn off most server-side data-fetching as a temporary workaround for bug #1282


richardolsson:
Superseded by #1284 which I think is a better fix. Please review. 😊

PR #1284, opened 2021-06-14
Preload data from correct org

### LINKED ISSUE #798: Switching organization does not affect preloading
This only affects users who have an official role in more than one organization. Switching organizations to any other than the default (first) one, and then attempting to navigate straight to (or refresh) any other section than the dashboard, will preload that section's content for the default organization. Steps to reproduce:

1. Navigate to Organize as a user who has official roles in more than one organization, making sure that the default organization ("A") is selected.
2. Switch to an organization "B" other than the default one
3. Navigate to the campaign section (action calendar)
4. Refresh the browser

Expected results: The content for B has been preloaded
Actual results: The content for A has been preloaded

To fix this, make sure that the route for switching organizations is executed before the preloader starts preloading (but after it creates the store).

richardolsson:
This PR fixes #798 by changing the order in which various middleware handle a request on the server-side. Before it went like this:

1. Set up store
2. Preload data
3. Set correct org from cookie

With this change, the order is now:

1. Set up store
2. Set correct org from cookie
3. Preload data

### DIFF
--- a/src/server/app.js
+++ b/src/server/app.js
@@ -19,7 +19,6 @@ import ActivistPage from '../components/fullpages/ActivistPage';
 import IntlReduxProvider from '../components/IntlReduxProvider';
 import ServerErrorPage from "../components/ServerErrorPage";
 import { setPanesFromUrlPath } from '../actions/view';
-import { setActiveOrg } from '../actions/user';
 
 const packageJson = require('../../../package.json');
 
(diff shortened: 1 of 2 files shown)


niklasva82:
Great! Seems to solve the bug!

niklasva82 (approved):

PR #1285, opened 2021-06-21
Undocumented/change bulk api delete

niklasva82:
Change bulk API delete to remove connection rather than delete person.


richardolsson (approved):
🚀

PR #1287, opened 2021-08-11
Workaround for server-client inconsistency with actions

### LINKED ISSUE #1286: Server/client inconsistency with actions at same time
It seems that when two actions begin at the exact same time, they might sometimes be ordered differently on the server and the client, causing render inconsistencies and exceptions to be thrown.
image

```
main.js:97 Server/client inconsistency near:
window.onload @ main.js:97
main.js:98 Client:  pan class="location" data-reactid="263">Laksevåg senter</span><span class="time"
window.onload @ main.js:98
main.js:99 Server:  pan class="location" data-reactid="263">Støletorget v/ 7-11</span><span class="t
window.onload @ main.js:99
main.js:100 Full HTML available in window.serverHtml and window.clientHtml
window.onload @ main.js:100
main.js:101 Invariant Violation: Minified React error #42; visit http://facebook.github.io/react/docs/error-decoder.html?invariant=42&args[]=%20(client)%20%3D%22%22%20data-reactid%3D%221%22%3E%3Chead%20data-reactid%3D%0A%20(server)%20%3D%22%22%20data-reactid%3D%221%22%20class%3D%22wf-frankling for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at reactProdInvariant (https://organize.zetk.in/static/main.js:6:49440)
    at Object._mountImageIntoNode (https://organize.zetk.in/static/main.js:31:208494)
    at mountComponentIntoNode (https://organize.zetk.in/static/main.js:31:201213)
    at ReactReconcileTransaction.perform (https://organize.zetk.in/static/main.js:17:215570)
    at batchedMountComponentIntoNode (https://organize.zetk.in/static/main.js:31:201527)
    at ReactDefaultBatchingStrategyTransaction.perform (https://organize.zetk.in/static/main.js:17:215570)
    at Object.batchedUpdates (https://organize.zetk.in/static/main.js:31:193360)
    at Object.batchedUpdates (https://organize.zetk.in/static/main.js:17:16950)
    at Object._renderNewRootComponent (https://organize.zetk.in/static/main.js:31:204741)
    at Object._renderSubtreeIntoContainer (https://organize.zetk.in/static/main.js:31:206892)
window.onload @ main.js:101
```

niklasva82:
Resolves #1286

This is a workaround for the server-client inconsistency for actions. It proved difficult to fix, so this PR simply removes preloading of actions on server render. The consequences of this may be a slight delay for displaying actions on reload or direct url entry of e.g. /campaign/actions on slower connections.

### DIFF
--- a/src/server/datarouter.js
+++ b/src/server/datarouter.js
@@ -2,7 +2,7 @@ import express from 'express';
 
 import { createLocalizeHandler } from './locale';
 import { configureStore } from '../store';
-import { retrieveActions, retrieveAction, retrieveActionsOnDay } from '../actions/action';
+import { retrieveAction, retrieveActionsOnDay } from '../actions/action';
 import { retrieveActionParticipants } from '../actions/participant';
 import { retrieveActionResponses } from '../actions/actionResponse';
 import { retrieveActivities } from '../actions/activity';
@@ -102,7 +102,6 @@ export default messages => {
     ]));
 
     router.get(/campaign\/locations$/, waitForActions(req => [
-        retrieveActions(),
         retrieveCampaigns(),
     ]));
 
@@ -115,20 +114,17 @@ export default messages => {
     ]));
 
     router.get(/campaign\/playback$/, waitForActions(req => [
-        retrieveActions(),
         retrieveActivities(),
         retrieveCampaigns(),
         retrieveLocations()
     ]));
 
     router.get(/campaign\/distribution$/, waitForActions(req => [
-        retrieveActions(),
         retrieveActivities(),
         retrieveCampaigns()
     ]));
 
     router.get(/campaign\/actions$/, waitForActions(req => [
-        retrieveActions(),
         retrieveActivities(),
         retrieveCampaigns(),
     ]));


richardolsson:
Were you able to reproduce the original bug? What fixes were tried and found not to work?

niklasva82:
Yes I was. The fixes tried are in issue-1286/server-client-inconsistency, I decided to create a new branch for the workaround.
Basically I tried sorting using id as a fallback in the reducer in order to make sure that the props given to the component was always the same. I tried sorting in the component in an earlier try and that didn't work. It did seem to work, but it caused other problems when creating new actions for instance, so I tried to implement an optional sort function argument to the updateOrAddListItems and createList.

When I couldn't get that to work without putting in too much work I decided it was better to work around the issue especially given that the current organize will be replaced, the performance penalty is not large and only applies to the specific situation of reloading or specifically going to pages with actions, e.g. /campaign/actions.

richardolsson (approved):
> […] I decided it was better to work around the issue especially given that the current organize will be replaced, the performance penalty is not large and only applies to the specific situation of reloading or specifically entering the URL of pages with actions, e.g. /campaign/actions.

Point taken! 😊 

Thanks for clarifying though. It's great for the documentation and mutual learning.

You can go ahead and merge this. 🎉

niklasva82:
Still can't merge
bild

PR #1289, opened 2021-08-23
Redirect ?org requests to current path, not /

### LINKED ISSUE #1288: Allow `org` query parameter on any path
The `org` query parameter should be allowed on any path, so that a user can switch organization without navigating to the home page.

This is mainly necessary for the MVP of Gen3, see zetkin/app.zetkin.org#218.

niklasva82:
Resolves #1288


richardolsson (approved):
So simple. Awesome! 💯

PR #1301, opened 2022-03-14
Fix select tag from specific organization bug

### LINKED ISSUE #1300: Bug: Select tag in specific organization
There is a bug in the tag select for specific organizations.

Steps to reproduce:
1. Create a new smart search
2. Add a tag filter
3. Select one of the organizations
4. Add a tag

Expected results:
The tag select pane should appear with only tags from the specified organization.

Actual results:
Nothing happens

niklasva82:
Resolves #1300


richardolsson (approved):
Looks good! I looked in the `filterByOrg` utility function to confirm that the correct name for this config variable is `specificOrganizations`.

PR #1303, opened 2022-03-15
Fix bug where anestors queries are not available to suborg query filter

### LINKED ISSUE #1302: Ancestor queries not included in suborg filter component
Steps to reproduce:
1. Create a smart search query in organization 1 and make it accessible to sub-organizations.
2. Switch to organization 4
3. Create a new smart search query
4. Select the sub-query filter

Expected results:
The query in organization 1 should be available as an option

Actual results:
The query in organization 1 is excluded even though it is returned from the API.

niklasva82:
Resolves #1302

### DIFF
--- a/src/utils/filterByOrg.js
+++ b/src/utils/filterByOrg.js
@@ -1,7 +1,7 @@
 export default (orgList, collection, config) => {
     let filterOrgs;
     if(config.organizationOption == 'all') {
-        filterOrgs = orgList;
+        return collection;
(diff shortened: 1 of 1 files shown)


richardolsson (approved):
This looks ok to me, but it's kinda hard to tell for me what's going on.

In the future, I think it would be great to include in the PR a one or two-sentence explanation to why/how it fixes a bug. In this case I'm just going to trust that it does, but it's difficult for me when I don't have a full understanding of the system to understand why a certain change fixes a bug.

PR #1305, opened 2022-03-22
Fix bug where survey filters sometimes crash.

niklasva82:
This fixes a bug that occasionally appears, especially with complex queries, where the survey related filters would crash due to the lack (presumably because it has not been properly loaded yet) of survey.organization.


richardolsson (approved):
LGTM! I missed this before the release, but will make an update with this in it now.

PR #1309, opened 2022-06-15
Implement caller query filter

niklasva82:
Screenshot from 2022-06-15 19-44-40
Screenshot from 2022-06-15 19-45-22


niklasva82:
Refactored to work with the refactored filter @richardolsson

niklasva82:
Refactored again

richardolsson (approved):
LGTM!

PR #1311, opened 2022-06-16
Fix query survey option bug

### LINKED ISSUE #1310: Query survey options bug
It appears my bugfix in the last release #1305 introduced a new bug where the survey with its elements is never loaded because the list of surveys has not been loaded.

### To reproduce:
1. Create a query with a survey option filter.
2. Reload the page and look at the query

#### Expected results:
The selected suryey option should be visible and editable.

#### Actual results: 
The survey options select box is not visible. The selected option can be found again by selecting another survey and then selecting the pre-selected survey again.

Screenshot from 2022-06-16 11-27-12

niklasva82:
Fixes #1310

Serious bug, this should be pushed as soon as possible IMO!

### DIFF
--- a/src/components/filters/SurveyOptionFilter.jsx
+++ b/src/components/filters/SurveyOptionFilter.jsx
@@ -30,7 +30,8 @@ export default class SurveyOptionFilter extends FilterBase {
     componentWillReceiveProps(nextProps) {
         this.setState(this.stateFromProps(nextProps));
 
-        if (nextProps.config.survey && nextProps.config.survey != this.props.config.survey) {
+        if (nextProps.config.survey && nextProps.config.survey != this.props.config.survey 
+                || !nextProps.elementsBySurvey[nextProps.config.survey]) {
             this.props.dispatch(retrieveSurvey(this.state.survey));
         }
     }


src/components/filters/SurveyOptionFilter.jsx
```diff
@@ -30,7 +30,8 @@ export default class SurveyOptionFilter extends FilterBase {
     componentWillReceiveProps(nextProps) {
         this.setState(this.stateFromProps(nextProps));
 
-        if (nextProps.config.survey && nextProps.config.survey != this.props.config.survey) {
+        if (nextProps.config.survey && nextProps.config.survey != this.props.config.survey 
+                || !nextProps.elementsBySurvey[nextProps.config.survey]) {
             this.props.dispatch(retrieveSurvey(this.state.survey));
```
richardolsson, line 35: The filter framework is really messy. I can't wait to move to Gen3. When I'm looking at this, I'm realizing that all of the conditions look at `nextProps` and `this.props`, but then when the action to retrieve the survey is dispatched, it uses  `this.state` to get the survey.

I guess that works if the `setState()` call on line 31 is synchronous, but I'm not sure it's guaranteed to be. Anyway, the change you've made makes sense!

richardolsson (approved):
This looks good to me. I will try to get it out in a patch release.

Is there any workaround at all? Does closing and opening the pane make a difference?

PR #1313, opened 2022-07-04
Fix bug where old all filters were missing a config object. 

niklasva82:
Also fix a bug in the AllCallAssignmentsPane related to recursive fetching, don't filter on the organization for now and assume that all call assignments are fetched as normal.


richardolsson (approved):
Looks good!

PR #1314, opened 2022-07-27
Add participants to the action export excel sheet

### LINKED ISSUE #1308: Action schedule printouts
Users have requested a schedule printout for actions that will list their participants and their contact information. This is especially important for multi-shift actions, so that the activists can always find the information about the person who will be relieving them.

niklasva82:
Resolves #1308 

I'm not quite satisfied with the format of the resulting Excel-sheet, but it works. The problem is that it doesn't fit on a single printed page. Another solution could be to do a sub-section between each action row with a header for the participant contact info, but it seems that formatting is int the "pro"-version of the xlsx-package so I'm not sure how to make it more presentable and clear.

Screenshot from 2022-07-28 01-29-04

### DIFF
--- a/src/server/api/bulk.js
+++ b/src/server/api/bulk.js
@@ -52,12 +52,104 @@ let operations = {
             'Contact name',
             'Contact e-mail',
             'Contact phone',
+            'Participant name',
+            'Participant e-mail',
+            'Participant phone',
         ];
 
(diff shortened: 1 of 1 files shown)


richardolsson (approved):
I checked this with the user who requested this feature, and they were happy! 💯

PR #1318, opened 2022-11-12
Survey options view column

niklasva82:
This appears to work but is untested with multiple options and the styling is not great. It also concatenates all (unique) options from all submissions, I don't know if this is the best choice or not. Needs some further work.


niklasva82:
Screenshot from 2022-11-12 02-39-01

niklasva82:
Screenshot from 2022-11-16 16-52-16

niklasva82:
Also sorted the selected options so that they are always in the same order, but maybe this is guaranteed anyways, I'm not sure?

richardolsson (approved):
LGTM! 💯

PR #1321, opened 2022-11-29
Fix view sorting for several column types

niklasva82:
Implement view sorting for the following column types:
`survey_option`
`survey_options`
`person_notes`
`local_person`
`survey_response`
`survey_submitted`


richardolsson (approved):
Great work! I think there's something to be improved to the way sorting handles empty cells, but I don't think it's worth struggling with in Gen2 since sorting is already so much more flexible in Gen3. I'm going to deploy this!

PR #1324, opened 2023-04-20
Set published to yesterday at 00:00 when creating a new action

### LINKED ISSUE #1322: Old reference to `GoogleAnalytics.jsx`
There is an old reference to `GoogleAnalytics.jsx` in `ActivistPage.jsx` causing `master` to crash. The reference should be deleted.

### LINKED ISSUE #1323: Set published to current datetime when creating an action
With the new published feature in gen3, the API is also changing. We need to set `published` to current datetime when creating an action in order for gen2 actions to keep working as expected.

niklasva82:
Resolves #1323 
Resolves #1322

### DIFF
--- a/src/actions/action.js
+++ b/src/actions/action.js
@@ -100,6 +100,13 @@ export function updateAction(id, data) {
 export function createAction(campaignId, data) {
     return ({ dispatch, getState, z }) => {
         let orgId = getState().org.activeId;
+
+        // This is a fix to make sure all created actions are set to be published yesterday
+        let yesterday = new Date();
+        yesterday.setHours(0,0,0,0);
+        yesterday.setDate(yesterday.getDate()-1);
+
+        data.published = yesterday;
         dispatch({
             type: types.CREATE_ACTION,
             meta: { campaignId },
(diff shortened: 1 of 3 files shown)


src/actions/action.js
```diff
@@ -100,6 +100,13 @@ export function updateAction(id, data) {
 export function createAction(campaignId, data) {
     return ({ dispatch, getState, z }) => {
         let orgId = getState().org.activeId;
+
+        // This is a fix to make sure all created actions are set to be published yesterday
+        let yesterday = new Date();
+        yesterday.setHours(0,0,0,0);
+        yesterday.setDate(yesterday.getDate()-1);
+
+        data.published = yesterday;
```
richardolsson, line 109: Wouldn't it make sense to do `yesterday.toISOString()` here, so that it's a serialized value and not a `Date` object?

src/actions/actionImport.js
```diff
@@ -96,6 +101,7 @@ export function executeActionImport(campaignId) {
                         num_participants_required: row.parsed.participants,
                         info_text: row.parsed.info,
                         title: row.parsed.title,
+                        published: yesterday,
```
richardolsson, line 104: Same here (see previous comment).

richardolsson (changes requested):
Do you reckon it would make sense to use serialized dates before sending it to the server, so that they're not implicitly serialized in a potentially incorrect format?

richardolsson (approved):
LGTM now!

PR #1325, opened 2023-04-25
Fix sorting bug for tags caused by changes in API

niklasva82:
Fixes sorting bug caused by new tag view column which includes values for value tags.

To reproduce bug:
1. Select a view
2. Add a tag column
3. Assign a tag to one of the people in the view
4. Click the sort button in the header of the tag column

Expected:
Column should sort on the presence of the specified tag.

Actual:
Nothing happens.

### DIFF
--- a/src/components/misc/personViews/PersonViewTable.jsx
+++ b/src/components/misc/personViews/PersonViewTable.jsx
@@ -139,9 +139,12 @@ export default class PersonViewTable extends React.Component {
 
                             let x = 0;
 
-                            if (colType == 'local_bool' || colType == 'person_tag' || colType == 'person_query') {
+                            if (colType == 'local_bool' || colType == 'person_query') {
                                 // Treat boolean values as integers (1 or 0)
                                 x = +val1 - +val0;
+                            }
+                            else if (colType == 'person_tag') {
+                                x = !!val0 < !!val1 ? 1 : -1;
                             }
                             else if (colType == 'survey_option') {
                                 // Find whether option has been selected in any of the submissions and treat bool as integer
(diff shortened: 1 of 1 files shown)


src/components/misc/personViews/PersonViewTable.jsx
```diff
@@ -162,7 +165,7 @@ export default class PersonViewTable extends React.Component {
                                 const joinedOptions0 = latest0.selected.reduce(joinText, '');
                                 const joinedOptions1 = latest1.selected.reduce(joinText, '');
                                 x = joinedOptions0.localeCompare(joinedOptions1);
-                            }
+                            }
```
richardolsson, line 168: I'm not happy about little things like this, but I'm glad we're avoiding them in app.zetkin.org thanks to linting and auto-formatting. 😊

src/components/misc/personViews/PersonViewTable.jsx
```diff
@@ -139,10 +139,13 @@ export default class PersonViewTable extends React.Component {
 
                             let x = 0;
 
-                            if (colType == 'local_bool' || colType == 'person_tag' || colType == 'person_query') {
+                            if (colType == 'local_bool' || colType == 'person_query') {
                                 // Treat boolean values as integers (1 or 0)
                                 x = +val1 - +val0;
                             }
+                            else if (colType == 'person_tag') {
+                                x = !!val0 < !!val1 ? 1 : -1;
```
richardolsson, line 147: Nice solution!

richardolsson (approved):
This looks good to me!

niklasva82:
Haha forgot that I had already solved this :smile:

PR #1328, opened 2023-05-22
Change max length of external to 96 char in validation

### LINKED ISSUE #1326: External ID as 36 char uuid in import
I have been trying to work out how to sync from Action Network to Zetkin using the people upload feature. Action Network lets you download CSV files of people on the system, but the ID is a string with letters numbers and hyphens (i.e. a UUID).

Not sure exactly what is needed for this, or how big of a change, but it would be great to have this as a simpler system for syncing

joe-irving:
* Changes max length of external ID from 12 to 96 in import validater
* Changes Swedish and English warning messages from 12 to 96

Resolves #1326 on front end, @richardolsson  to double check that back end is okay for longer external IDs

### DIFF
--- a/locale/panes/confirmImport/en.yaml
+++ b/locale/panes/confirmImport/en.yaml
@@ -9,7 +9,7 @@ dateInvalid: Date invalid on rows { rows }.
 duplicatesInfo: You have selected duplicate columns.
 emailInvalid: Invalid E-mail address on rows { rows }.
 errorTitle: Invalid data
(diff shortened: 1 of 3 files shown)


richardolsson:
@niklasva82 can you review this, and make sure it works with backend?

niklasva82 (approved):
Looks good! Verified that import of up to 96 character external ids works and verified that the same ID is returned by the API.

PR #1330, opened 2023-09-15
Change map from Google to Leaflet (OSM)

niklasva82:
* Replace Google Maps with Leaflet for LocationMap and AddressMap
* Remove static maps because no simple and public reliable static map generator for OSM.
* Remove Campaign playback heatmap. There were a few alternatives, but quite old.


richardolsson (changes requested):
Everything looks great, except one thing that I discovered in the normal location map (http://organize.dev.zetkin.org/maps). Clicking a marker there doesn't show the pane. It opens it but it seems to be behind the map. Basically it's a z-index issue, so hopefully it should be easy to fix?

richardolsson (approved):
This seems to be working as expected now! 💯

PR #1331, opened 2023-09-30
Move JS build to runtime from docker build

richardolsson:
This PR moves the building of frontend code to runtime, instead of the docker build. This is so that a single image built from the code in this repo can be used in multiple environments.


richardolsson:
Approved offline. Merging!

PR #1333, opened 2024-06-13
Fix gen 3-related bugs

niklasva82:
This fixes some gen 3 related bugs where the action activity, campaign and location can be null. It also resolves an issue with the organize backend not starting up properly resulting in a gateway timeout when trying to access it in devmode.

image
image


richardolsson (approved):
This looks good to me!

